Issue #57, opened 2020-04-19
Add filter_enum support for checkboxset control


matthoward366:

matthoward366:
Done.

PR #58, opened 2020-04-19
Add support for checkboxset enum filtering

matthoward366:

           in the file; or

           it.

For issue #57


canvas_modules/common-canvas/__tests__/common-properties/controls/checkboxset-test.js
```diff
@@ -205,7 +205,7 @@ describe("checkboxset control tests", () => {
 		// unchecked a box
 		checkboxes.at(2).getDOMNode().checked = true;
 		checkboxes.at(2).simulate("change");
-		expect(controller.getPropertyValue(propertyIdInvalid)).to.eql(["apple", "orange", "peach"]);
+		expect(controller.getPropertyValue(propertyIdInvalid)).to.eql(["orange", "peach"]);
```
matthoward366, line 208: This was actually working incorrectly.  Invalid values should be filtered out.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -898,6 +899,22 @@ export default class PropertiesController {
 		if (!skipValidateInput) {
 			conditionsUtil.validateInput(inPropertyId, this);
 		}
+		// check enum filters and filter out any invalid values
+		const propertyValue = this.getPropertyValue(propertyId);
```
matthoward366, line 903: This was added to filter out values that can get added by default.  This shows up in tables when we add default rows of values and sometimes those values should be filtered.

canvas_modules/common-canvas/__tests__/test_resources/paramDefs/checkboxset_paramDef.json
```diff
@@ -263,6 +299,31 @@
...
+        }
+      },
+      {
+        "parameter_ref": "checkboxset_filtered",
+        "label": {
+          "default": "checkboxset Filtered"
+        },
+        "resource_key": "checkboxset_filtered",
+        "description": {
+          "default": "checkboxset filtered to have values 'red' and 'blue'"
```
caritaou, line 324: This should be `'apple' and 'pear'`
image

canvas_modules/common-canvas/src/common-properties/properties-controller.js
caritaou: I see when we click `Add Value`, all the existing rows get updated to the new filtered list. Should the filter only apply to new rows added?

Similarly, if I enable the filter on a table and click to remove a row, the rows are not updated to the filtered values until I click `Add Value`. Following the same pattern when adding, should the rows be updated when removing a row?

matthoward366: I was debating on the best way of handling this.  The checkboxset test is a little odd.  Based on the condition we probably should be updating the values in the table and not just the new rows since users wouldn't be able to update existing rows.

I'll fix this so it works for both add and remove.  I think in most cases the filter wouldn't change once the values are added.

canvas_modules/common-canvas/__tests__/test_resources/paramDefs/checkboxset_paramDef.json
matthoward366: good catch.  I'll get this fixed.

matthoward366:
```
Can you update the documentation http://localhost:3001/#/conditions#--filteredEnum to include checkboxset in the list of support controls?
```
I'm not seeing where we list supported controls for this.  Maybe I'm missing it.

canvas_modules/common-canvas/src/common-properties/controls/abstract-table.jsx
```diff
@@ -320,11 +320,10 @@ export default class AbstractTable extends React.Component {
 			} else {
 				cellContent = (<div className="properties-table-cell-control">
 					<ReadonlyControl
-						control={this.props.control}
+						control={this.props.controller.getControl(propertyId)}
```
matthoward366, line 323: This was passing in the parent control when we should have been passing in the actual control to be render.  We no longer need columnDef because of that.

canvas_modules/common-canvas/src/common-properties/controls/structurelisteditor/structurelisteditor.jsx
```diff
@@ -33,8 +33,7 @@ class StructurelisteditorControl extends AbstractTable {
 
 	addRow() {
 		const newRow = this.getDefaultRow();
-		const rows = this.props.controller.getPropertyValue(this.props.propertyId);
-		rows.push(newRow);
+		const rows = this.props.controller.getPropertyValue(this.props.propertyId).concat([newRow]);
```
matthoward366, line 36: This is so we get a new property value instead of updating the reference value which causes issues in updatePropertyValue.

PR #80, opened 2020-05-11
Implemented nodes unit test in Cypress

nmgokhale:

           in the file; or

           it.

### DIFF
--- a/canvas_modules/harness/cypress/support/canvas/node-cmds.js
+++ b/canvas_modules/harness/cypress/support/canvas/node-cmds.js
@@ -211,6 +211,18 @@ Cypress.Commands.add("deleteNode", (nodeLabel) => {
 	cy.verifyNodeIsDeleted(nodeLabel);
 });
 
+Cypress.Commands.add("deleteNodeUsingKeyboard", (nodeName) => {
+	// Delete node by pressing 'Delete' key on keyboard
+	cy.useDeleteKey()
+		.then((deleteKey) => {
+			cy.getNodeForLabel(nodeName)
+				.click()
+				.type(deleteKey);
+			// Verify node is deleted
+			cy.verifyNodeIsDeleted(nodeName, true);
+		});
+});
+
 Cypress.Commands.add("getNodeDimensions", (nodeLabel) => {
 	cy.getNodeForLabel(nodeLabel).then((node) => {
 		const nodeDimensions = {
@@ -221,4 +233,26 @@ Cypress.Commands.add("getNodeDimensions", (nodeLabel) => {
 		};
 		return nodeDimensions;
 	});
+});
+
+Cypress.Commands.add("selectAllNodes", () => {
+	cy.get("#canvas-div-0").find(".node-image")
+		.then((nodes) => {
+			cy.useShiftKey()
+				.then((shiftKey) => {
+					// Press and hold the shift key
+					cy.get("body")
+						.type(shiftKey, { release: false });
+
+					// Click all the nodes
+					nodes.each((idx, node) => {
+						cy.wrap(node)
+							.click();
+					});
+
+					// Cancel the shift key press
+					cy.get("body")
+						.type(shiftKey, { release: true });
+				});
+		});
 });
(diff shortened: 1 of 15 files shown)


canvas_modules/harness/cypress/support/canvas/node-cmds.js
```diff
@@ -14,9 +14,16 @@
  * limitations under the License.
  */
 
-Cypress.Commands.add("getNodeForLabel", (nodeLabel) =>
-	cy.get(getNodeGrpSelector())
-		.then((grpArray) => findGrpForLabel(grpArray, nodeLabel)));
+Cypress.Commands.add("getNodeForLabel", (nodeLabel) => {
```
tomlyn, line 17: It's only trivial but could you change getNodeForLabel to be getNodeWithLabel  so it uses the same terminology as getCommentWithText ?

canvas_modules/harness/cypress/support/canvas/node-cmds.js
```diff
@@ -212,3 +235,25 @@ Cypress.Commands.add("getNodeDimensions", (nodeLabel) => {
...
+						cy.wrap(node)
+							.click();
+					});
+
+					// Cancel the shift key press
+					cy.get("body")
+						.type(shiftKey, { release: true });
+				});
+		});
+});
```
tomlyn, line 259: I just realized this not strictly correct. The shift key is supposed to be used to select a range of nodes in a flow from a start to end node and all those in between.  To select all the nodes on a canvas the user should press the Cmnd/Ctrl key and then click each node in turn.  The reason this works is that as we select each node with shift it does not deselect and previously selected nodes -- which, to be honest, I was surprised by when I tested it since I think it should. So this behavior may change in the future.

So I think this function ought to be changed to be using the Cmnd/Ctrl key instead of Shift.

canvas_modules/harness/cypress/support/canvas/object-model-cmds.js
```diff
@@ -49,6 +49,34 @@ Cypress.Commands.add("getSupernodePipeline", (supernodeName) => {
 	});
 });
 
+Cypress.Commands.add("getNodeFromObjectModel", (nodeName) => {
```
tomlyn, line 52: Can you change this to getNodeCountFromObjectModel ?

canvas_modules/harness/cypress/support/canvas/object-model-cmds.js
```diff
@@ -49,6 +49,34 @@ Cypress.Commands.add("getSupernodePipeline", (supernodeName) => {
...
+			nodes.forEach(function(node) {
+				if (node.label === nodeName) {
+					count++;
+				}
+			});
+			return count;
+		});
+});
+
+Cypress.Commands.add("getCommentFromObjectModel", (commentText) => {
```
tomlyn, line 66: Can you change this to getCommentCountFromObjectModel ?

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
```diff
@@ -86,6 +109,19 @@ Cypress.Commands.add("verifyCommentIsNotSelected", (commentText) => {
 	cy.isCommentSelected(commentText).should("eq", false);
 });
 
+Cypress.Commands.add("verifyCommentIsAdded", (commentText) => {
```
tomlyn, line 112: Let's change this to `verifyCommentExists`  because it might be used to check if, say, comments exist in a canvas that was loaded. Also change the corresponding method name for nodes and links if they exist.

canvas_modules/harness/cypress/support/canvas/object-model-cmds.js
```diff
@@ -90,3 +140,10 @@ Cypress.Commands.add("clearMessagesFromAllNodes", () => {
 			});
 	});
 });
+
+Cypress.Commands.add("isObjectModelEmpty", () => {
```
tomlyn, line 144: We'd usually expect a command beginning with `is` to return a boolean whereas this is returning a count so I think the name should be `getObjectCountFromObjectModel`

Issue #104, opened 2020-06-01
Update notification center


aqtang:
New design:

- Able to open notification center when empty
- styling changes
- subtitle, close button, and clear all button added

![image](https://user-images.githubusercontent.com/10677021/83438525-692c1380-a3f6-11ea-8d4f-e22e18a31b6a.png)

Issue #25, opened 2020-04-02
Close the notifications panel only when the notifications icon in the toolbar is clicked


avigadmi:
When notification panel is open we display more details about node errors and we don't want this info to disappear on every click. We would like to have a new config flag that will allow us to close the notifications panel only when the notifications icon in the toolbar is clicked. 

![image (9)](https://user-images.githubusercontent.com/36592011/78219813-fb806b80-74c8-11ea-95b6-59e9c8b65c37.png)

aqtang:
This feature has been added in:
https://github.com/elyra-ai/canvas/commit/d1fb5ac6766ad72d902fb6555978d9fca6246191

There is now a boolean `keepOpen` that can be added to the `notificationConfig` object passed to `<CommonCanvas>` (which then goes to `<NotificationPanel>`), which if true will keep the notification center open until it's closed directly

tomlyn:
Implemented in 8.0.23

PR #105, opened 2020-06-01
Update notification center

aqtang:
Fixes:
- https://github.com/elyra-ai/canvas/issues/25
- https://github.com/elyra-ai/canvas/issues/104

New features:

- Optional clear all button
- Options to dismiss individual messages
- Spacing and structure for a subtitle
- Ability to open notification center when empty, with option to pass in a default empty message
- Option to keep notification center open when clicking elsewhere
- A close button
- Styling changes
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
+++ b/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
@@ -495,7 +495,7 @@ export default class CanvasController {
 		this.objectModel.clearNotificationMessages();
 	}
 
-	// Removes the notification message(s) with the given id(s)
+	// Removes the notification messages from the given array of ids
 	deleteNotificationMessages(ids) {
 		this.objectModel.deleteNotificationMessages(ids);
 	}
--- a/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
+++ b/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
@@ -495,6 +495,11 @@ export default class CanvasController {
 		this.objectModel.clearNotificationMessages();
 	}
 
+	// Removes the notification message(s) with the given id(s)
+	deleteNotificationMessages(ids) {
+		this.objectModel.deleteNotificationMessages(ids);
+	}
+
 	// Returns the array of currently displayed notification messages shown in
 	// the notification panel. The format of a notification message is an object
 	// with these fields:
--- a/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
+++ b/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
@@ -149,6 +149,19 @@ class NotificationPanel extends React.Component {
 					{this.props.notificationConfig && this.props.notificationConfig.emptyMessage ? this.props.notificationConfig.emptyMessage : null}
 				</div>
 			</div>);
+		const clearAll = this.props.notificationConfig && this.props.notificationConfig.clearAllMessage
+			? (<div className="notification-panel-clear-all-container">
+				<Button
+					className="notification-panel-clear-all"
+					onClick={this.clearNotificationMessages.bind(this)}
+					kind="ghost"
+					size="small"
+					disabled={this.props.messages.length === 0}
+				>
+					{this.props.notificationConfig.clearAllMessage}
+				</Button>
+			</div>)
+			: null;
 
 		return (<div className={"notification-panel-container " + notificationPanelClassName} >
 			<div className="notification-panel">
@@ -161,17 +174,7 @@ class NotificationPanel extends React.Component {
 					<div className="notification-panel-messages">
 						{notificationPanelMessages}
 					</div>
-					<div className="notification-panel-clear-all-container">
-						<Button
-							className="notification-panel-clear-all"
-							onClick={this.clearNotificationMessages.bind(this)}
-							kind="ghost"
-							size="small"
-							disabled={this.props.messages.length === 0}
-						>
-							{this.props.notificationConfig ? this.props.notificationConfig.clearAllMessage : null}
-						</Button>
-					</div>
+					{clearAll}
 				</div>
 				<svg className="notification-popup-arrow" x="0px" y="0px" viewBox="0 0 16 9">
 					<polyline points="0,9 8,0 16,9" />
--- a/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
+++ b/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
@@ -120,8 +120,8 @@ class NotificationPanel extends React.Component {
 		}
 	}
 
-	deleteNotification(messageKey) {
-		this.props.canvasController.deleteNotificationMessages(messageKey);
+	deleteNotification(id) {
+		this.props.canvasController.deleteNotificationMessages(id);
 	}
 
 	clearNotificationMessages() {
--- a/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
+++ b/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
@@ -99,16 +99,18 @@ class NotificationPanel extends React.Component {
 	}
 
 	handleNotificationPanelClickOutside(e) {
-		const notificationIcon = document.getElementsByClassName("notificationCounterIcon")[0];
-		const notificationHeader = document.getElementsByClassName("notification-panel-header")[0];
-		const notificationMessages = document.getElementsByClassName("notification-panel-messages-container")[0];
+		if (!this.props.notificationConfig.keepOpen) {
+			const notificationIcon = document.getElementsByClassName("notificationCounterIcon")[0];
+			const notificationHeader = document.getElementsByClassName("notification-panel-header")[0];
+			const notificationMessages = document.getElementsByClassName("notification-panel-messages-container")[0];
 
-		if (!this.props.notificationConfig.keepOpen && this.props.isNotificationOpen &&
-				notificationIcon && !notificationIcon.contains(e.target) &&
-				notificationHeader && !notificationHeader.contains(e.target) &&
-				notificationMessages && !notificationMessages.contains(e.target)) {
-			this.props.canvasController.toolbarActionHandler("closeNotificationPanel");
-			e.stopPropagation(); // Prevent D3 canvas code from clearing the selections.
+			if (this.props.isNotificationOpen &&
+					notificationIcon && !notificationIcon.contains(e.target) &&
+					notificationHeader && !notificationHeader.contains(e.target) &&
+					notificationMessages && !notificationMessages.contains(e.target)) {
+				this.props.canvasController.toolbarActionHandler("closeNotificationPanel");
+				e.stopPropagation(); // Prevent D3 canvas code from clearing the selections.
+			}
 		}
 	}
 
--- a/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
+++ b/canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
@@ -59,7 +59,7 @@ class NotificationPanel extends React.Component {
 			</div>);
 
 			const closeMessage = message.closeMessage
-				? (<div className = "notification-message-close" onClick={this.deleteNotification.bind(this, message.key)}>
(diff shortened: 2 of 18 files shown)


aqtang:
if no `clearAllMessage` is set in the notificationConfig, no "Clear all" button section will be rendered

image


if `keepOpen` is passed in and set to true in the notificationConfig, clicking outside the notification panel will not close it, it must be closed by clicking on the new close button or on the toolbar icon.

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
```diff
@@ -106,31 +120,87 @@ class NotificationPanel extends React.Component {
...
+					{notificationSubtitle}
+					<Close16 className="notification-panel-close-icon" onClick={this.closeNotificationPanel} />
+				</div>
+				<div className="notification-panel-messages-container">
+					<div className="notification-panel-messages">
+						{notificationPanelMessages}
+					</div>
+					{clearAll}
+				</div>
+				<svg className="notification-popup-arrow" x="0px" y="0px" viewBox="0 0 16 9">
```
caritaou, line 179: The arrow should be the same color as the title background
image

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
```diff
@@ -47,49 +49,89 @@ $notification-left-border-color-width: 4px;
...
-	max-height: calc((100% / 2) - #{$notification-panel-header-height});
-	overflow-y: auto;
+.notification-panel-clear-all-container {
+	width: 100%;
+	height: $notification-panel-clear-all-height;
 	background-color: $ui-01;
+	display: flex;
+	flex-direction: column;
+	justify-content: center;
 	border: 1px solid $ui-03;
```
caritaou, line 89: The border above the "Clear all" should be just 1px, its combined with the bottom border from messages right now
image

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
```diff
@@ -106,31 +120,87 @@ class NotificationPanel extends React.Component {
...
+				<div className="notification-panel-empty-message">
+					{this.props.notificationConfig && this.props.notificationConfig.emptyMessage ? this.props.notificationConfig.emptyMessage : null}
 				</div>
-				<svg className="notification-popup-arrow" x="0px" y="0px" viewBox="0 0 16 9">
-					<polyline points="0,9 8,0 16,9" />
-				</svg>
+			</div>);
+		const clearAll = this.props.notificationConfig && this.props.notificationConfig.clearAllMessage
+			? (<div className="notification-panel-clear-all-container">
+				<Button
```
caritaou, line 154: Should this entire area be a button? Users need to hover over the text in order to clear all messages right now

image

canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
```diff
@@ -495,6 +495,11 @@ export default class CanvasController {
 		this.objectModel.clearNotificationMessages();
 	}
 
+	// Removes the notification message(s) with the given id(s)
```
tomlyn, line 498: Can we mention that ids is an array of IDs ?
BTW - What I've done for CanvasController API documentation is to take the method prologs and method signature from the code (like this one) and copy them into the Wiki. So hopefully you can do the same with this method prolog and method signature.

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
```diff
@@ -106,31 +120,87 @@ class NotificationPanel extends React.Component {
 		}
 	}
 
+	deleteNotification(messageKey) {
+		this.props.canvasController.deleteNotificationMessages(messageKey);
+	}
```
tomlyn, line 125: This seems a little strange. The method name is singular but the method it's calling is plural with the same parmeter passed through and the parameter name is singular as well. Are these keys or IDs that are being passed here?  Is there one key (ID) or many?

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
```diff
@@ -16,6 +16,8 @@
 
 $notification-panel-arrow-height: 8px;
 $notification-panel-header-height: 35px;
+$notification-panel-clear-all-height: 49px;
+$notification-message-height: 64px;
```
tomlyn, line 20: This file has a strange mixture of the two scss variables above plus some 'hardcoded' pixel dimenions in the rulesets below (like notification-panel width).  Can we move any 'hardcoded' sizes into this set of variables?

canvas_modules/common-canvas/src/object-model/object-model.js
```diff
@@ -947,6 +947,13 @@ export default class ObjectModel {
 		return notificationMessages;
 	}
 
+	deleteNotificationMessages(ids) {
+		// accept a single key or array of keys
```
tomlyn, line 951: This is a great example of an unnecessary comment which has already not managed to keep up with changes to the code -- which is often what happens with comments in the code. :-)  I would remove it but if you want to keep it it should be corrected.

canvas_modules/common-canvas/src/object-model/object-model.js
```diff
@@ -947,6 +947,13 @@ export default class ObjectModel {
...
+		const filterIds = Array.isArray(ids) ? ids : [ids];
+		const newMessages = this.getNotificationMessages().filter((message) => !filterIds.includes(message.id));
+		this.store.dispatch({ type: "SET_NOTIFICATION_MESSAGES", data: newMessages });
+	}
```
tomlyn, line 955: It threw me for a minute when I saw the variables was called 'newMessages'.  Maybe 'retainedMessages' would be better.
I think I might have been tempted to add a new redux action called DELETE_NOTIFICATION_MESSAGES and do the work inside the body of the redux action - but I guess this will work.

canvas_modules/common-canvas/src/toolbar/toolbar.jsx
```diff
@@ -85,20 +85,20 @@ class Toolbar extends React.Component {
...
 			} else if (successMessages.length > 0) {
-				className = notificationIconClassName + constants.SUCCESS;
-			} else if (notificationMessages.length > 0) {
-				className = notificationIconClassName + constants.INFO;
-			} else {
-				className = notificationIconClassName;
+				className += constants.SUCCESS;
+			} else if (infoMessages.length > 0) {
+				className += constants.INFO;
 			}
```
tomlyn, line 102: I'm a bit confused by the changes above.  Looks like the code used to pick the class for the highest error level with the current messages (as the comment still suggests). Now it is concatenating class names together (without a space between them) -- is that correct?

canvas_modules/harness/cypress/integration/properties/custom-controls.js
```diff
@@ -22,6 +22,11 @@ describe("Test of custom panels", function() {
 	});
 
 	it("Test custom panels are working correctly", function() {
+		// this next click is a workaround, the next click on the toggle won't trigger
+		// an onClick unless elsewhere on the panel is clicked first. remove when fixed
+		cy.get(".harness-custom-control-custom-toggle").first()
+			.find(".bx--form-item")
+			.click();
 		cy.get(".harness-custom-control-custom-toggle label").first()
```
tomlyn, line 30: For code that possibly needs to future action can you add a TODO  string in the comment.

canvas_modules/harness/src/client/App.js
```diff
@@ -1641,8 +1641,10 @@ export default class App extends React.Component {
 			{ action: "arrangeVertically", label: "Arrange Vertically", enable: true }
 		];
 
-		const notificationConfig = { action: "notification", label: "Notifications", enable: true, notificationHeader: "Notifications" };
-		const notificationConfig2 = { action: "notification", label: "Notifications", enable: true, notificationHeader: "Notifications Canvas 2" };
+		const notificationConfig = { action: "notification", label: "Notifications", notificationHeader: "Notification Center", notificationSubtitle: "subtitle status", enable: true,
+			emptyMessage: "You don't have any notifications right now.", clearAllMessage: "Clear all", keepOpen: true };
+		const notificationConfig2 = { action: "notification", label: "Notifications", notificationHeader: "Notification Center Canvas 2", enable: true,
+			emptyMessage: "You don't have any notifications right now.", clearAllMessage: "Clear all" };
 		const contextMenuConfig = {
```
tomlyn, line 1648: I know I commented on it before but ... let's make sure we document these new config options in the Wiki.
BTW - How does the user indicate they don't want the Clear All button?  Is it by ommiting that field from the config?

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
aqtang: I'll check with Analusia

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
aqtang: fixed

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
aqtang: fixed

canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
aqtang: updated the description comment. I'll copy over the signature and explain in the wiki as well when merged

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
aqtang: The "key" I was generating is no longer used, it's just the message ID now, I've updated the parameter to be clearer.

In the notification panel itself, this function is used only for dismissing individual messages, hence the singular name. I wanted to make the higher level CanvasController function a bit more flexible in case we have other cases in the future, but in the panel itself, deleting individual messages and the clear all button covers all the use cases

canvas_modules/common-canvas/src/object-model/object-model.js
aqtang: removed, I agree this is unnecessary

canvas_modules/common-canvas/src/toolbar/toolbar.jsx
aqtang: if you look at line 92
```
			className += " ";
```

If an extra icon classname is needed, a space is added. When I originally wrote this section of the code, I was adding a space in each of the 4 if cases instead of just adding it once above.

The logic in this section changed a little bit because the notification center is no longer disabled even when empty, so there is no more disabled state

canvas_modules/harness/src/client/App.js
aqtang: Yes, if no "clear all" message is passed in, then no button will be shown. I will make sure to document this clearly in the wiki.

canvas_modules/common-canvas/src/object-model/object-model.js
aqtang: I see now the phrasing is a bit confusing. "newMessages" made sense to me since this new array of messages comprises the new entirety of the message log, though it is just a filtered, smaller version of the previous state.

This redux method/naming is also used for clearing all messages (either through the sidepanel api in the test harness or the new clear all button, so I thought it'd be more consistent this way.

canvas_modules/common-canvas/src/notification-panel/notification-panel.jsx
aqtang: Confirmed with her that this is the intended design

canvas_modules/harness/cypress/integration/properties/custom-controls.js
aqtang: added a TODO, thanks!

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
aqtang: `$notification-message-height` is used in several places and `$notification-panel-clear-all-height` is used in a calculation. the width was only used in once place, as you said, so that's the only new spacing change that I did not create a variable for

Issue #156, opened 2020-07-08
Upgrade node version to 12.18.2


aqtang:
This will fix newly discovered memory leak vulnerability

https://nodejs.org/en/blog/release/v12.18.2/

PR #157, opened 2020-07-08
#156 Upgrade node version to 12.18.2

aqtang:
Addresses https://github.com/elyra-ai/canvas/issues/156

Upgrade node version in travis
 
           in the file; or

           it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ branches:
     - master-last_success
     - release-*
 language: node_js
-node_js: "12"
+node_js: "12.18.2"
 addons:
   apt:
     packages:
@@ -28,7 +28,7 @@ addons:
       - unzip
 before_install:
   - . ./scripts/setEnv.sh
-  - npm install -g npm@5.6.0
+  - npm install -g npm@6.14.5
   - npm install -g grunt-cli
 services:
   - xvfb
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,15 +20,15 @@ branches:
...
+node_js: "12.18.2"
 addons:
   apt:
     packages:
       - google-chrome-stable
       - unzip
 before_install:
   - . ./scripts/setEnv.sh
-  - npm install -g npm@5.6.0
+  - npm install -g npm@6.14.5
```
matthoward366, line 31: We should just remove this and default to the node_js version.

.travis.yml
```diff
@@ -20,15 +20,15 @@ branches:
     - master-last_success
     - release-*
 language: node_js
-node_js: "12"
+node_js: "12.18.2"
```
matthoward366, line 23: We should keep this generic.  Does it use the latest if we have 12 or do we need 12.x?

aqtang: `12 latest 12.x release` 
I'll revert that, it will be generic enough 
12 latest 12.x release

.travis.yml
aqtang: removed

Issue #168, opened 2020-07-20
Ability to return parameter IDs in a complex structure control


caritaou:
Values returned from complex structures may be hard to identify which ID maps to what value. 

<img width="298" alt="image" src="https://user-images.githubusercontent.com/1628100/87968166-dbbf8400-ca74-11ea-8ec7-f9455821cd09.png">

We need a way to allow the parameter IDs to be returned with the values, for example:
```
"data": {
    "form": {
      "structuretableReadonlyColumnStartValue": [
        {
          "field": "Cholesterol",
          "structuretable_sort_order_readonly_int": 5,
          "structuretable_sort_order": "Ascending"
        },
        {
          "field": "BP",
          "structuretable_sort_order_readonly_int": 6,
          "structuretable_sort_order": "Ascending"
        }
      ]
    }
```

caritaou:
This is implemented with https://github.com/elyra-ai/canvas/pull/169 and https://github.com/elyra-ai/pipeline-schemas/pull/25

A new `type` property has been added to the complex type definition schema. This defaults to `array` and can be set to `object`. The parameters will be read in and passed out in this format defined.

PR #169, opened 2020-07-20
#168 structures to return objects of parameter ids and value pairs

caritaou:

           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/util/property-utils.js
+++ b/canvas_modules/common-canvas/src/common-properties/util/property-utils.js
@@ -15,7 +15,7 @@
  */
 
 import logger from "../../../utils/logger";
-import { ParamRole } from "../constants/form-constants";
+import { ParamRole, ControlType } from "../constants/form-constants";
 import { DATA_TYPE, CARBON_ICONS } from "../constants/constants";
 import { cloneDeep } from "lodash";
 import { v4 as uuid4 } from "uuid";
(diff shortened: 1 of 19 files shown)


canvas_modules/common-canvas/src/common-properties/util/property-utils.js
```diff
@@ -183,6 +183,72 @@ function convertType(storage) {
...
+ *  Example currentValues: [[{a: 1, b: 2}], [{a: 10, b; 20}]]
+ *  Example convertedValues: [[1, 2], [10, 20]]
+ */
+function convertObjectStructureToArray(controlType, subControls, currentValues) {
+	const structureKeys = [];
+	subControls.forEach((control) => {
+		structureKeys.push(control.name);
+	});
+
+	if (controlType === ControlType.STRUCTUREEDITOR) {
```
matthoward366, line 198: Why do we need to check control type?  The control really shouldn't matter.  It should be base on the type.

canvas_modules/common-canvas/src/common-properties/util/property-utils.js
```diff
@@ -183,6 +183,72 @@ function convertType(storage) {
...
+ *  Example currentValues: [[1, 2], [10, 20]]
+ *  Example convertedValues: [[{a: 1, b: 2}], [{a: 10, b; 20}]]
+ */
+function convertArrayStructureToObject(controlType, subControls, currentValues) {
+	const structureKeys = [];
+	subControls.forEach((control) => {
+		structureKeys.push(control.name);
+	});
+
+	if (controlType === ControlType.STRUCTUREEDITOR) {
```
matthoward366, line 233: Same as above.

canvas_modules/common-canvas/src/common-properties/util/property-utils.js
caritaou: This function will only be called in the case where `type` equals `object`. This check is needed to differentiate the different currentValues received. `STRUCTUREEDITOR` is a single object since it doesn't have any rows whereas the other stcutures have 1 or many rows. I'll need to fix the example values in the comment

Issue #278, opened 2020-09-22
Add landmark roles for Accessibility


nmgokhale:
Add following landmark roles for accessibilty -
Common canvas header --> Banner
Palette --> Navigation
Canvas --> Main
Canvas Toolbar --> toolbar
Right side panel --> Complementary

PR #279, opened 2020-09-23
#278 Added landmark roles for A11y

nmgokhale:
Fixes: https://github.com/elyra-ai/canvas/issues/278
To test this PR -

1. Turn on VoiceOver (command + F5)
2. Open rotor (control + option + u)
3. Use left-right arrow keys to open Landmark menu
4. Note - Toolbar won't show up in landmark menu because `role=toolbar` is not a landmark role.

**Landmark menu with Flyout Palette -**
Image from iOS (6)

**Landmark menu with Modal Palette -**
Image from iOS (5)


tomlyn (changes requested):
Should we be putting the aria label text into the translation files?

nmgokhale:
> Should we be putting the aria label text into the translation files?

`Common Canvas Header banner `and `Right Side Panel complimentary` are a part of test harness. So translations are not added for these two roles. 
Other roles are translated - 
Screen Shot 2020-09-30 at 11 15 03 AM

**Toolbar -** 
canvas toolbar translation

canvas_modules/common-canvas/locales/palette/locales/en.json
```diff
@@ -1,3 +1,5 @@
 {
-  "palette.flyout.search.placeholder": "Find palette nodes"
+  "palette.flyout.search.placeholder": "Find palette nodes",
+  "palette.flyout.label": "Flyout Palette",
+  "palette.modal.label": "Modal Palette"
```
tomlyn, line 4: Can you change this to be "Dialog Palette" -- it sin't really a Modal and was badly named in the first place

tomlyn (approved):

Issue #316, opened 2020-10-11
Select All option (keyboard or context menu) should select links as well as nodes and comments


tomlyn:
The Select All option (keyboard or context menu) should select links as well as nodes and comments when the `enableLinkSelection` config value is set to something other than `None`

PR #317, opened 2020-10-11
#316 Select All option (keyboard or context menu) should select links…

tomlyn:
… as well as nodes and comments

Fixes: #316 

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/object-model/api-pipeline.js
+++ b/canvas_modules/common-canvas/src/object-model/api-pipeline.js
@@ -101,7 +101,7 @@ export default class APIPipeline {
 		return linksToDelete;
 	}
 
-	isEmpty(includeLinks) {
+	isEmpty() {
 		if (this.getNodes() && this.getNodes().length === 0 &&
 				this.getComments() && this.getComments().length === 0 &&
 				this.getLinks() && this.getLinks().length === 0) {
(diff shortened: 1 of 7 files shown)


canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -101,9 +101,10 @@ export default class APIPipeline {
 		return linksToDelete;
 	}
 
-	isEmpty() {
+	isEmpty(includeLinks) {
 		if (this.getNodes() && this.getNodes().length === 0 &&
-				this.getComments() && this.getComments().length === 0) {
+				this.getComments() && this.getComments().length === 0 &&
+				this.getLinks() && this.getLinks().length === 0) {
```
caritaou, line 107: Should `getLinks()` be called only if `includeLinks` is true?

tomlyn: Hi Carita - good catch. Actually I introduced the `includeLinks` param because I initially thought that this check needs to be dependent on whether detachable links is switched on or not. However, I realized that if link selection is switched off there cannot be any detached links without there being at last a couple of nodes/comments. So I removed the `includeLinks` param from the code but forgot to remove it from the function signature. (If you trace back through the code it is not actually passed in from here in common-canvas.jsx:
```js
isCanvasEmpty={this.canvasController.isPrimaryPipelineEmpty()}
```

caritaou (approved):

Issue #319, opened 2020-10-13
A11y: Add landmark role for Common-properties


nmgokhale:
Role name: Properties
Landmark Role: Complementary

PR #340, opened 2020-10-22
#319 Add landmark role for common-properties

nmgokhale:
Fixes: https://github.com/elyra-ai/canvas/issues/319
To test this PR -

1. Turn on VoiceOver (command + F5)
2. Open rotor (control + option + u)
3. Use left-right arrow keys to open Landmark menu
Properties Landmark Role

           in the file; or

           it.


matthoward366 (changes requested):
Are we able to put this just in the render method of common-properties
```
render() {
		if (this.state.hasError) {
			return this.displayAppropriateUIFallback();
		}
		return this.createPropertiesMainComponent();
	}
```
Instead of returning each of these set a variable then return the aside with the content.  That would handle all cases.

nmgokhale:
Adding `<aside />` in render method of common-properties doesn't capture buttons in the properties modal. 
Landmark role is added only for `Flyout` view. 
[Modal view has its own `role=dialog`](https://www.carbondesignsystem.com/components/modal/accessibility/#how-it-works). Usually modals are not added in landmark roles.
Screen Shot 2020-10-22 at 1 36 22 PM

matthoward366 (approved):
Looks good.

Issue #436, opened 2021-01-06
Add flag to persist properties editor size


marthacryan:
When the width of a properties right-flyout editor is changed, the width doesn't persist to new editors.

For example:
* open a properties editor - the default size is small
* resize editor to medium
* close the editor
* open a new properties editor - it will open in size small

There doesn't appear to be a clean way to persist the size to the next editor you open, because while you can use the DOM to check the current editor's size and use the `uihints.editorSize` field to open the new editor in that size, the resize button is set up as follows:
* If the default size is small, resize toggles between small and medium
* If the default size is medium, resize toggles between medium and large

So when you manually set the default size based on the info in the DOM, you get this situation:

* Open a properties editor - the default size is small
* resize editor to medium
* close the editor, record that the size of the editor is medium
* open a new properties editor and manually set the default size to medium
* press the resize button, it toggles the size to large, you can't toggle the size to small anymore

Could we add a flag, maybe something along the lines of `persistEditorSize`, that uses the width of the last editor that was opened instead of always using the default? 

This is the code that causes the issue - it'd just need a little added logic checking if the `persistEditorSize` flag is set to true before setting the size to the default:
https://github.com/elyra-ai/canvas/blob/030265f8fc3e12bbc09b0f148a72adbdd136d756/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx#L91-L94

PR #486, opened 2021-02-16
 #436 Persist properties editor size

nmgokhale:
**Please do not merge**

Does not persist editor size -
- When I click on `Close` button and reopen the list_paramDef.json
- When I select any other paramDef.json from dropdown list

Persists editor size only in this case -
persist-editor-size

 
           in the file; or

           it.


canvas_modules/common-canvas/src/common-properties/common-properties.jsx
```diff
@@ -73,6 +73,7 @@ class CommonProperties extends React.Component {
 				callbacks= {this.props.callbacks}
 				customControls={this.props.customControls}
 				customConditionOps={this.props.customConditionOps}
+				currentEditorSize={this.props.currentEditorSize}
```
matthoward366, line 76: I would move this under `propertiesInfo` and use `initialEditorSize`

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -47,6 +47,10 @@ class PropertiesMain extends React.Component {
 	constructor(props) {
 		super(props);
 		this.propertiesController = new PropertiesController();
+		// Persist editorSize when resize() is not called
+		if (this.propertiesController.currentEditorSize !== this.props.currentEditorSize) {
```
matthoward366, line 51: Why check this?  I would expect to only set this if something is set in `this.props.currentEditorSize`

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -350,20 +354,24 @@ class PropertiesMain extends React.Component {
 				this.setState({
 					editorSize: Size.MEDIUM
 				});
+				this.propertiesController.setCurrentEditorSize(Size.MEDIUM);
```
matthoward366, line 357: I would move the setting of the editor size to a method and do the state and controller change together.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -96,6 +96,10 @@ export default class PropertiesController {
 		return this.commandStack;
 	}
 
+	setCurrentEditorSize(currentEditorSize) {
```
matthoward366, line 99: We should have a get command.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -47,6 +47,10 @@ class PropertiesMain extends React.Component {
 	constructor(props) {
 		super(props);
 		this.propertiesController = new PropertiesController();
+		// Persist editorSize when resize() is not called
+		if (this.propertiesController.getInitialEditorSize() !== this.props.propertiesInfo.initialEditorSize) {
```
nmgokhale, line 51: This condition is required to persist editorSize when resize() is not called. 
If I remove this condition and go through following steps, editorSize is not persisted -
1. I open `list_paramDef.json` and click on `Resize` button --> Size is updated to Medium
2. Then I open `numberfield_paramDef.json` --> Opening size is Medium (Persisted editorSize from step 1). 
3. I don't click on `Resize` button in step 2
4. I open `list_paramDef.json` --> Opening size is Small which is default size (editorSize is not persisted)

persist-editor-size

matthoward366: Why not check if `this.props.propertiesInfo.initialEditorSize` is set and set it?  Not sure why we need to check the old value.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -96,6 +96,14 @@ export default class PropertiesController {
 		return this.commandStack;
 	}
 
+	setInitialEditorSize(initialEditorSize) {
```
matthoward366, line 99: Thinking about this more this probably makes more sense to rename to setEditorSize and getEditorSize.  Sorry about that.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -168,6 +174,20 @@ class PropertiesMain extends React.Component {
 		return PropertyUtils.formatMessage(this.props.intl, MESSAGE_KEYS.PROPERTIESEDIT_REJECTBUTTON_LABEL);
 	}
 
+	getEditorSize() {
+		// Determine whether to persist initialEditorSize or set the defaultEditorSize in certain cases
+		const defaultEditorSize = this.propertiesController.getForm().editorSize;
```
matthoward366, line 179: Do you know how this works if someone uses the pixel size instead of small, medium, large?

nmgokhale: If `editor_size` is not specified in the paramDef.json, default `edior_size` is set to `small`. And rightFlyoutPanel has width set to `pixel_width.min`.


For example -
`Filter_paramDef.json` doesn't have `editor_size`. But it has `pixel_width: { "min": 200, "max": 700 }`.
If persisted editor_size is `small`, right flyout panel has `width: 200px`
```
<aside aria-label="Properties" role="complementary" class="properties-wrapper properties-right-flyout properties-small" tabindex="0" style="width: 200px;">
```
If persisted editor_size is `medium`, right flyout panel has `width: 700px`
```
<aside aria-label="Properties" role="complementary" class="properties-wrapper properties-right-flyout properties-medium" tabindex="0" style="width: 700px;">
```
If persisted editor_size is `large`, default editor_size is set to `small` and right flyout panel has `width: 200px`
```
<aside aria-label="Properties" role="complementary" class="properties-wrapper properties-right-flyout properties-small" tabindex="0" style="width: 200px;">
```

canvas_modules/harness/src/client/App.js
```diff
@@ -1242,10 +1242,13 @@ class App extends React.Component {
 			if (currentEditorNodeId && canvasController.getNode(currentEditorNodeId, activePipelineId)) {
 				commonPropertiesRef.applyPropertiesEditing(false);
 			}
+			let initialEditorSize;
```
matthoward366, line 1245: Instead of always passing this initialEditorSize in can we add a dropdown in the properties configuration panel that allows us to test with and without setting this?  I would create it with unset, small, medium, large.  When unset we shouldn't pass anything into the propertyInfo object.

matthoward366, line 1245: Instead of always passing this initialEditorSize in can we add a dropdown in the properties configuration panel that allows us to test with and without setting this?  I would create it with unset, small, medium, large.  When unset we shouldn't pass anything into the propertyInfo object.

matthoward366 (approved):
Changes look good.

Issue #557, opened 2021-03-12
Support enable/disable move row buttons in table


nmgokhale:
Add a new method to the properties controller that can be used to disable all move row buttons. 
We'll still enable/disable them using the current rules in the control.

<img width="600" alt="Move-row-buttons" src="https://user-images.githubusercontent.com/25124000/110873524-cce94400-8286-11eb-9aba-13c4808ab8a9.png">

PR #558, opened 2021-03-12
#557 Support disable move row buttons in table

nmgokhale:
Fixes https://github.com/elyra-ai/canvas/issues/557
 
           in the file; or

           it.


canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -750,6 +751,21 @@ export default class PropertiesController {
...
 
+	/**
+	 * Enable/disable table row move buttons
+	 * enableTableRowMoveButtons() doesn't enable all the buttons. Buttons will be enabled based on row selection
+	 * @param propertyId Id of the property for the control to operate upon
+	 *
+	 */
+	disableTableRowMoveButtons(inPropertyId) {
+		const propertyId = this.convertPropertyId(inPropertyId);
+		this.disableMoveRowButtons[propertyId.name] = true;
```
matthoward366, line 762: This needs to be stored using propertyId so embedded controls can disable/enable the buttons.  This also needs to be part of redux so we can listen to changes and rerender the control on changes.

matthoward366 (changes requested):
How did you test this?  I think we need a way in the test harness to validate this.

canvas_modules/common-canvas/src/common-properties/components/moveable-table-rows/moveable-table-rows.jsx
```diff
@@ -230,5 +244,13 @@ MoveableTableRows.propTypes = {
...
 	setScrollToRow: PropTypes.func.isRequired,
 	tableContainer: PropTypes.object.isRequired,
-	disabled: PropTypes.bool
+	disabled: PropTypes.bool,
+	disableRowMoveButtonsForPropertyIds: PropTypes.array // set by redux
 };
+
+const mapStateToProps = (state, ownProps) => ({
+	// Array of propertyIds - propertyId used for disabling row move buttons
+	disableRowMoveButtonsForPropertyIds: state.disableRowMoveButtonsReducer.propertyIdArray
```
matthoward366, line 253: It would be better to have a method in the propertyController to get the state for a specific property.  This saves on rerendering the react component.  Having the entire array means each change to that array rerender every component listening to that array.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -750,6 +750,15 @@ export default class PropertiesController {
 		this.propertiesStore.clearSelectedRows(propertyId);
 	}
 
+	/**
+	 * Disable table row move buttons for all propertyIds in given array
+	 * @param propertyIdArray Array of propertyIds
+	 *
+	 */
+	disableRowMoveButtons(propertyIdArray) {
```
matthoward366, line 758: It looks like this is an array of strings.  Any reason you are using an array over a single property?  I'm guessing for a single property we'd have to allow for enabling it.  

Users should be passing in an object for the propertyId.  Similar to what we do with all other controller methods.  We can store it in redux as a string though.

matthoward366 (changes requested):

canvas_modules/common-canvas/src/common-properties/properties-controller.js
nmgokhale: I am using an array, so host applications can add multiple propertyIds in array. And move row buttons will be disabled for all propertyIds. 
Example -` [{"name":"list_string"}, {"name":"list_string_error"}] `

nmgokhale:
This code works for nested tables if the correct propertyId is passed to the `setDisableRowMoveButtons()` in controller.

For example -
Screen Shot 2021-03-18 at 2 34 22 PM

Image shows parent table. Nested table opens when we click on edit button in each row.
Parent table has propertyId - `{"name": "nestedStructuretable"}`
Nested table has propertyId - `{"name": "nestedStructuretable", "row": 0, "col": 3}`

- If we pass parent table's propertyId  - `setDisableRowMoveButtons([{"name": "nestedStructuretable"}])` , move row buttons in parent table will be disabled. Nested table is not affected.

- If we pass nested table's propertyId - `setDisableRowMoveButtons([{"name": "nestedStructuretable", "row": 0, "col": 3}])` , move row buttons in nested table will be disabled. Parent table is not affected.

- If we pass both propertyIds - `setDisableRowMoveButtons([{"name": "nestedStructuretable"}, {"name": "nestedStructuretable", "row": 0, "col": 3}])` , move row buttons will be disabled in both the tables.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -756,6 +757,40 @@ export default class PropertiesController {
...
+	getDisableRowMoveButtons() {
+		return this.propertiesStore.getDisableRowMoveButtons();
+	}
+
+	/**
+	 * Check if row move buttons should be disabled for given propertyId
+	 * @param propertyId  The unique property identifier
+	 * @return boolean
+	 */
+	checkIfDisableRowMoveButtons(propertyId) {
```
matthoward366, line 785: I would probably rename this to `isDisableRowMoveButtons`

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -756,6 +757,40 @@ export default class PropertiesController {
...
+		const propertyIds = this.getDisableRowMoveButtons();
+		return (
+			Array.isArray(propertyIds)
+				? propertyIds.some((el) => getDataId(el) === getDataId(propertyId))
```
matthoward366, line 789: I wonder if using https://lodash.com/docs/4.17.15#isEqual would work better here?  Then you don't have to worry about converting to a string.

matthoward366 (commented):
A couple comment but everything looks good.

matthoward366 (approved):
Looks good.

matthoward366 (approved):

Issue #503, opened 2021-02-23
The config option enableHightlightNodeOnNewLinkDrag has a typo


tomlyn:

tomlyn:
Fixed

PR #569, opened 2021-03-19
#568 Allow nodes to be disabled when a new or existing link is being …

tomlyn:
…dragged

Fixes: #568 

 
           in the file; or

           it.


canvas_modules/common-canvas/src/common-canvas/svg-canvas-d3.js
```diff
@@ -90,6 +90,7 @@ export default class SVGCanvasD3 {
 				this.config.enableDragWithoutSelect !== config.enableDragWithoutSelect ||
 				this.config.enableParentClass !== config.enableParentClass ||
 				this.config.enableHightlightNodeOnNewLinkDrag !== config.enableHightlightNodeOnNewLinkDrag ||
```
caritaou, line 92: Should this be renamed as well? I see https://github.com/elyra-ai/canvas/issues/503 has been fixed and you're updating the examples in this PR

canvas_modules/common-canvas/src/common-canvas/svg-canvas-renderer.js
```diff
@@ -3330,48 +3376,21 @@ export default class SVGCanvasRenderer {
...
 	setImageContent(imageSel, image) {
-		const nodeImageType = this.getImageType(image);
 		if (image !== imageSel.attr("data-image")) {
+			const nodeImageType = this.getImageType(image);
 			// Save image field in DOM object to avoid unnecessary image refreshes.
 			imageSel.attr("data-image", image);
 			if (nodeImageType === "svg") {
-				this.addCopyOfImageToDefs(image);
-				this.updateUseObject(imageSel, image);
+				d3.xml(image, { cache: "force-cache" }).then((data) => {
```
caritaou, line 3385: Are the images still stored in `defs`?

tomlyn: No unfortunately with that approach we can only style icons once no matter how many times they are referenced from nodes. This means if we had two nodes both referring to the same icon we couldn't style one with fill = "grey" and one that is not grey. So now I include the SVG in the DOM for each node so in the example above the SVG would be inserted twice; once for each node. This means the icons can be styles separately on  a node by node basis.

canvas_modules/common-canvas/src/common-canvas/svg-canvas-d3.js
tomlyn: Yes, good catch. I found some other places with that typo. None will cause any runtime error but I fixed them nonetheless.

caritaou (approved):

Issue #609, opened 2021-04-01
Add configuration settings to have condition logic treat hidden/disabled properties as null or false


matthoward366:

matthoward366:
These options have been added
```
conditionHiddenPropertyHandling: PropTypes.oneOf(["null", "value"]),
conditionDisabledPropertyHandling: PropTypes.oneOf(["null", "value"])
```
It will be available in the next elyra canvas build.  I'll also update the documentation once published.

matthoward366:
Done

PR #610, opened 2021-04-01
#609 condition handling config options

matthoward366:

           in the file; or

           it.

#609


canvas_modules/common-canvas/__tests__/common-properties/operators/equalsOp-test.js
```diff
@@ -55,7 +55,7 @@ describe("validating equals operator works correctly", () => {
 		expect(equals(wrap(null), wrap(null), null, controller)).to.equal(true);
 		expect(equals(wrap({ temp: "value" }), wrap({ temp: "value2" }), null, controller)).to.equal(false);
 		// pass in a function as a way to hit the default switch case
-		expect(equals(wrap(emptyFunc), null, null, controller)).to.equal(true);
+		expect(equals(wrap(emptyFunc), wrap({ temp: "value2" }), null, controller)).to.equal(true);
```
matthoward366, line 58: Extra logic was added to operation to handle nulls.  This is needed to hit the default condition.

canvas_modules/common-canvas/__tests__/common-properties/operators/equalsOp-test.js
```diff
@@ -69,7 +69,7 @@ describe("validating equals operator works correctly", () => {
 		expect(equals(wrap(null), undefinedPlaceholder, null, controller)).to.equal(true);
 		expect(equals(wrap({ temp: "value" }), undefinedPlaceholder, { temp: "value2" }, controller)).to.equal(false);
 		// pass in a function as a way to hit the default switch case
-		expect(equals(wrap(emptyFunc), undefinedPlaceholder, null, controller)).to.equal(true);
+		expect(equals(wrap(emptyFunc), undefinedPlaceholder, "value", controller)).to.equal(true);
```
matthoward366, line 72: Extra logic was added to operation to handle nulls.  This is needed to hit the default condition.

canvas_modules/common-canvas/src/common-properties/ui-conditions/condition-ops/equals.js
```diff
@@ -23,7 +23,10 @@ function op() {
 function evaluate(paramInfo, param2Info, value, controller) {
 	if (paramInfo.control.controlType !== "passwordfield") {
 		const dataType = typeof paramInfo.value;
-		if (typeof param2Info !== "undefined") {
+		if (param2Info) {
+			if (paramInfo.value === null || param2Info.value === null) {
```
caritaou, line 27: With this condition defined, it will never get into the `object` case. Should this be moved there instead?

canvas_modules/common-canvas/src/common-properties/ui-conditions/condition-ops/notEquals.js
```diff
@@ -22,7 +22,10 @@ function op() {
 function evaluate(paramInfo, param2Info, value, controller) {
 	if (paramInfo.control.controlType !== "passwordfield") {
 		const dataType = typeof paramInfo.value;
-		if (typeof param2Info !== "undefined") {
+		if (param2Info) {
+			if (paramInfo.value === null || param2Info.value === null) {
```
caritaou, line 26: Same question as above

caritaou (changes requested):
I found some problems when testing these changes with the panelConditions_paramDef.json, specifically when disabling/hiding nested panels. After clicking `Hide 'Level1' panel`, the checkbox became disabled and it cannot be clicked again.
image

canvas_modules/common-canvas/src/common-properties/ui-conditions/condition-ops/equals.js
matthoward366: I added this here because if the cases where the param2Info.value === null but paramInfo.value=string.  It would drop into the string case and fail.  I removed the check from object since it will never be hit.

matthoward366:
For `panelConditions_paramDef.json` this is because the condition that enables the button is looking for false and not null.  I could had some extra handling in the conditions to enable if false or null.

caritaou (approved):

PR #631, opened 2021-04-15
#630 Palette search string including spaces adds extra characters

tomlyn:
Fixes: #630 
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
+++ b/canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
@@ -28,7 +28,7 @@ export default (state = {}, action) => {
 	case "REMOVE_NODE_TYPES_FROM_PALETTE":
 	case "SET_CATEGORY_LOADING_TEXT":
 	case "SET_CATEGORY_EMPTY_TEXT": {
-		return Object.assign({}, state, { categories: categories(state.categories, action) });
+		return Object.assign({}, state, { categories: categories(state.categories, action), version: "3.0" });
 	}
 	default:
 		return state;
--- a/canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
+++ b/canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
@@ -28,6 +28,7 @@ export default (state = {}, action) => {
 	case "REMOVE_NODE_TYPES_FROM_PALETTE":
 	case "SET_CATEGORY_LOADING_TEXT":
 	case "SET_CATEGORY_EMPTY_TEXT": {
+		// Save version for the palette in case this is the first action performed on an empty palette.
 		return Object.assign({}, state, { categories: categories(state.categories, action), version: "3.0" });
 	}
 	default:
--- a/canvas_modules/common-canvas/src/palette/palette-flyout-utils.js
+++ b/canvas_modules/common-canvas/src/palette/palette-flyout-utils.js
@@ -26,7 +26,7 @@ export function getFilteredNodeTypeInfos(categories, searchString) {
 		filteredNodeTypeInfos.push(...getFilteredNodeTypeInfosByCategory(categories[idx], filterStrings));
 	}
 	const rankedFilteredNodeTypeInfos =
-		filteredNodeTypeInfos.sort((e1, e2) => ((e1.occurenceInfo.ranking <= e2.occurenceInfo.ranking) ? 1 : -1));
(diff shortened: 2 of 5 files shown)


tomlyn:
@caritaou 
* I changed the ranking algorithm to give more weighting to an occurrence of a keyword in the node label.
I switched off the tooltip for search results because the tooltip just shows what is in the palette search result item (Vlad wanted that)
I fixed the issue Neha found that causes text to be repeated in the displayed text when you enter overlapping search strings
I also fixed a few other minor miscellaneous items that I’d found over the last few days

tomlyn:
@caritaou 
I made the following changes:
* I fixed the issue Neha found that causes text to be repeated in the displayed text when you enter overlapping search strings
* I changed the ranking algorithm to give more weighting to an occurrence of a keyword in the node label.
* I switched off the tooltip for search results because the tooltip just shows what is in the palette search result item (Vlad wanted that)
* I also fixed a few other minor miscellaneous items that I’d found over the last few days
*  I changed the spelling of occurrence which I'd spelt as occurence

canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
```diff
@@ -28,7 +28,7 @@ export default (state = {}, action) => {
 	case "REMOVE_NODE_TYPES_FROM_PALETTE":
 	case "SET_CATEGORY_LOADING_TEXT":
 	case "SET_CATEGORY_EMPTY_TEXT": {
-		return Object.assign({}, state, { categories: categories(state.categories, action) });
+		return Object.assign({}, state, { categories: categories(state.categories, action), version: "3.0" });
```
caritaou, line 31: Why are you saving the version here?

canvas_modules/common-canvas/src/palette/palette-flyout-utils.js
```diff
@@ -37,150 +37,173 @@ function getFilteredNodeTypeInfosByCategory(category, filterStrings) {
...
-	const addOccurences = [];
-	newOccurences.forEach((newOcc) => {
+	const addOccurrences = [];
+	newOccurrences.forEach((newOcc) => {
 		let handled = false;
-		occurences.forEach((occ) => {
-			if (newOcc.start >= occ.start && newOcc.start <= occ.end) {
-				if (newOcc.end > occ.end) {
+		occurrences.forEach((occ) => {
+			//   Occurrence         <-->
```
caritaou, line 144: I dont understand these comments, can you elaborate? Or are the comments required here?

canvas_modules/common-canvas/src/object-model/redux/reducers/palette.js
tomlyn: Typically, when a palette is loaded into common-canvas it is has the `version: "3.0"` property set or, if it is an older version, common-canvas upgrades it to v3.0 and then loads it. v1.0 of the palette schema did not have a version property defined so the absence of that property should indicate a v1.0 palette.

When using the 'save to palette' option on a node I noticed that, if there is no current palette loaded, the action would create a palette in Redux correctly with a new category and the node in it, but the in-memory palette would not have a version property.  

This means that, if the palette was saved to a file and then later loaded into common-canvas, the absence of a version number would make common-canvas think that the palette was a v1.0 palette (because we didn't have the version property for v1.0). Common-canvas would then have problems loading that palette because the absence of the version property should indicate the palette is v1.0 but the actual contents (category and node) comply with v3.0. Consequently, I added the version property in the line above to avoid that problem.

canvas_modules/common-canvas/src/palette/palette-flyout-utils.js
tomlyn: The algorithm is joining two arrays . Each array contains elements which describe an 'occurrence' of one of the filter strings in the main string being analyzed that is, either the category label, the node label or the description. The algorithm loops through the arrays looking for overlaps between occurrences so it can consolidate them together. So if, for example, I had and occurrence of { start: 3, end: 5 } and another of { start: 4, end: 7 } they would get consolidated into a single occurrence of  { start: 3, end: 7 }.

There are a few different combinations of this type of overlap which the `if` statements within the loop try to handle. The comments try to show the different possibilities of overlap that the associated `if` statement is handling.

I noticed actually that in globally correcting the spelling of `occurrence` it nudged the lines slightly out of place. So I have corrected that and added an additional line to show that the arrows are describing string occurrences. I'd like to keep the comments here even if they only make sense to me :)

caritaou (approved):

Issue #652, opened 2021-05-13
Update Notification panel CSS


nmgokhale:
Currently Notification panel looks like this -
![Screen Shot 2021-05-13 at 11 24 15 AM](https://user-images.githubusercontent.com/25124000/118169425-c6d62780-b3dd-11eb-8ffe-379c9366cc3d.png)

After styling changes, notification panel should look like this -
- Remove arrow and spacing between toolbar and notification panel
- Header should use typography `$productive-heading-02`
- Toolbar button shows blue bottom-border when notification panel is open
![Screen Shot 2021-05-13 at 2 59 42 PM](https://user-images.githubusercontent.com/25124000/118193062-ff84f980-b3fb-11eb-913f-8c2936b1a3fc.png)

Toolbar buttons - 
When toolbar button is focused, show `$hover-field` background color.
![canvas-toolbar-button-focus](https://user-images.githubusercontent.com/25124000/118193257-4115a480-b3fc-11eb-8b83-986317dfced2.gif)

PR #653, opened 2021-05-13
#652 Updated notification panel styling

nmgokhale:
Fixes https://github.com/elyra-ai/canvas/issues/652
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/__tests__/common-canvas/common-canvas-toolbar-test.js
+++ b/canvas_modules/common-canvas/__tests__/common-canvas/common-canvas-toolbar-test.js
@@ -280,6 +280,45 @@ describe("Common Canvas Toolbar renders correctly with config as OBJECT", () => 
 		expect(wrapper.find(".toolbar-item.toggleNotificationPanel-action")).to.have.length(0);
 	});
 
+	// When notification panel is open, toolbar button should have class notification-panel-selected
+	// this class shows blue bottom border for the button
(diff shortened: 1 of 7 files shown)


tomlyn (changes requested):
Hi Neha — It looks good. The only comment I’d make is that I wanted, if possible, for the code in the common toolbar to be independent of the idea of a notification panel.  In effect, thinking of the common-toolbar and common-canvas as two separate components where common-toolbar provides the functionality to have a button which has an associated panel. 

So in the common-toolbar code, instead of talking about a ‘notification panel’, we’d call it ‘associated panel’ or something.  The idea being that applications that want to use common-toolbar could add a button to the toolbar which displays an ‘associated  panel’  and the app would be able to put whatever they want in the panel.
   
In the case of common-canvas (driven by modeler flows) it would be notifications but applications could add whatever contents to the associated panel that they want.

The reason I wanted it this way is we have had apps dev teams ask for this ability.

Following this design, this line of code
```
"notification-panel-selected": actionObj.action === TOOLBAR_ACTIONS.TOOLBAR_TOGGLE_NOTIFICATION_PANEL && this.props.isNotificationOpen },
```
would not fit in because the display of the associated panel would not be for the ‘notification panel action’ but for any button for which an associated panel was defined. 

For the name of the panel, ‘associated panel’ is OK but I would ask Devin if there is a name being used by design to describe this type of panel.  

If we are short of time we can go with the changes you've made and perhaps do what I am suggesting later.

tomlyn (approved):

Issue #657, opened 2021-05-21
Enable editable decoration labels for links and nodes


JuanKUYang:
Some applications want to support the feature for links just as node-- you hover over a link name, see the pencil, click to edit.

PR #660, opened 2021-06-04
#657 Enable editable decoration labels for links and nodes

tomlyn:
Fixes: #657 
 
           in the file; or

           it.


canvas_modules/common-canvas/src/common-canvas/svg-canvas-utils-decs.js
```diff
@@ -0,0 +1,235 @@
...
+	getDec(id, decorations) {
+		if (decorations) {
+			const dec = decorations.find((nd) => nd.id === id);
+			return dec;
+		}
+		return null;
+	}
+
+	getDecLabelForeignClass(dec) {
+		const outlineClass = dec.label_outline ? " d3-node-label-outline" : "";
```
caritaou, line 34: Should this have a different class name from node? This is adding a border for this label, should it be bigger so the edit icon can fit inside or should there be no borders for link labels?

image

tomlyn: Hi @caritaou Good catch, however the label outline is just for debug purposes which is why I didn't document the `label_outline` property in the wiki. It's just so I can see the extent of the div. If the host app wants an outline around the label they would need to specify a `shape` decoration in addition to the label decoration.

caritaou (approved):

PR #756, opened 2021-08-18
#755 Support conditional defaults in common-properties

### LINKED ISSUE #755: Add support for conditional defaults in common-properties
A new conditional will be required to set default values on controls based on the `current_parameters` and conditional logic.

- When condition is true, set control value to Default value. 
- Prioritize condition default value over parameter.default value.
- We still honor "regular" default values if no conditional ones apply
- Order matters when evaluating defaults. If multiple conditions evaluate to true only the first one is used.

Here is a potential example
```
{
  "default_value": {
    "parameter_ref": "updated_param",
    "value": "New default value",
    "evaluate": {
      "condition": {
        "parameter_ref": "some_param",
        "op": "isNotEmpty"
      }
    }
  }
}
```

**Set default value flowchart -** 
Default value flowchart

nmgokhale:
Fixes #755 

Updated `defaults_paramDef.json` in test harness -
Screen Shot 2021-09-03 at 10 17 29 AM

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/properties-controller.js
+++ b/canvas_modules/common-canvas/src/common-properties/properties-controller.js
@@ -154,16 +154,27 @@ export default class PropertiesController {
 			}
 			// Set the opening dataset(s), during which multiples are flattened and compound names generated if necessary
 			this.setDatasetMetadata(datasetMetadata);
-			// For each control, check if Conditional default is present
-			controls.forEach((control) => {
-				const propertyId = { name: control.name };
-				const defaultValue = conditionsUtil.setConditionalDefaultValue(propertyId, this);
-			});
 
 			this.setPropertyValues(propertyValues); // needs to be after setDatasetMetadata to run conditions
 			// for control.type of structuretable that do not use FieldPicker, we need to add to
 			// the controlValue any missing data model fields.  We need to do it here so that
 			// validate can run against the added fields
+
+			// Evaluate conditional defaults based on current_parameters upon the loading of view
+			if (!isEmpty(propertyValues)) {
+				Object.keys(propertyValues).forEach((propertyName) => {
+					const propertyId = { name: propertyName };
+					const conditionalDefaultValue = conditionsUtil.getConditionalDefaultValue(propertyId, this);
+					if (
+						conditionalDefaultValue !== null &&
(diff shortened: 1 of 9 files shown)


canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -1011,6 +1040,23 @@ export default class PropertiesController {
 			this.propertiesStore.updatePropertyValue(propertyId, initialValue);
 			return;
 		}
+		// evaluate conditional defaults upon each control input change
+		if (this.form) {
+			let propertyValues = {};
```
matthoward366, line 1045: Why do we need this along with the 161-182 above?  They seem to be doing the same thing.

canvas_modules/common-canvas/src/common-properties/ui-conditions/conditions-utils.js
```diff
@@ -330,6 +330,46 @@ function allowConditions(inPropertyId, controller) {
...
+				const parameterRef = validation.definition.default_value.parameter_ref;
+				//  For a given parameter_ref, only the first default_value condition is used.
+				if (!(parameterRef in conditionalDefaultValues)) {
+					const result = UiConditions.validateInput(validation.definition, inPropertyId, controller);
+					if (result) {
+						// Condition evaluate to true
+						conditionalDefaultValues[parameterRef] = { result: result, value: validation.definition.default_value.value };
+					} else {
+						// get the parameter default value
+						const parameterRefControl = controller.getControl({ name: parameterRef });
```
matthoward366, line 361: We already have code that sets default values.  Wouldn't this conflict with that?

matthoward366 (commented):
Code looks good.  Just a couple comments.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
nmgokhale: 161-182 - Called initially when properties are loaded. This is called only once. 

1043-1059 - Called after properties value is changed, check if condition is still true. This can be called multiple times.
For example - condition checks if `mode = Include`. If I update the value of `mode`, condition should return false.

matthoward366 (approved):
Looks good.

PR #891, opened 2021-12-14
#890 Added callback to allow Undo/Redo label override for Save properties action

### LINKED ISSUE #890: Add a callback to allow Undo/Redo label to be overriden by the host app.
Ref: 3040

Issue #888, implemented a mechanism for modifying the tooltip text for the Undo and Redo buttons in common-canvas so the text can explain what action is being undone or redone. This includes a new callback in common-canvas that can be used to override all the common-canvas action labels. This change has added a `getLabel()` method to the three standard methods currently supported by the Action class.  

Dec-10-2021 15-30-42

This issue covers the requirement for common-properties to provide a callback so that the action label, for the one action common-properties provides, can also be overridden by the host application. This needs to be implemented in the `commonPropertiesAction.js` program and surfaced to the host application in a way that is consistent with the way common-properties offers callbacks in its API.

I have added some dummy code into `commonPropertiesAction.js` with a TODO to indicate where the code needs to be added.

Note: You will need to have PR #889 merged before the common-properties part can be implemented.

nmgokhale:
Fixes #890 
Added callback method `propertiesActionLabelHandler()`.

Setting custom label for "Save properties" action in test harness -
Screen Shot 2021-12-14 at 2 14 21 PM

Screen Shot 2021-12-14 at 2 14 27 PM

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/command-actions/commonPropertiesAction.js
+++ b/canvas_modules/common-canvas/src/command-actions/commonPropertiesAction.js
@@ -14,14 +14,18 @@
  * limitations under the License.
  */
 import Action from "../command-stack/action.js";
+import defaultMessages from "../../locales/command-actions/locales/en.json";
 
 export default class CommonPropertiesAction extends Action {
-	constructor(newValues, initialValues, appData, applyPropertyChanges) {
+	constructor(newValues, initialValues, appData, applyPropertyChanges, propertiesActionLabelHandler, intl) {
 		super(newValues);
 		this.newValues = newValues;
 		this.initialValues = initialValues;
 		this.appData = appData;
 		this.applyPropertyChanges = applyPropertyChanges;
+		this.propertiesActionLabelHandler = propertiesActionLabelHandler;
+		this.intl = intl;
+		this.defaultMessages = defaultMessages;
 	}
 
 	// Standard methods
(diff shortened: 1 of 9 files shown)


canvas_modules/common-canvas/src/command-actions/commonPropertiesAction.js
```diff
@@ -37,9 +41,23 @@ export default class CommonPropertiesAction extends Action {
...
-	// file or call a getActionLabel callback.
 	getLabel() {
-		return "";
+		if (this.propertiesActionLabelHandler) {
+			const label = this.propertiesActionLabelHandler();
+			if (label) {
+				return label;
+			}
+		}
+		return this.getDefaultLabel("action.commonProperties");
```
caritaou, line 51: Since the translated text is part of `common-canvas/locales/command-actions/locales/en.json`, you dont need to pass in the `intl` object from common-properties. You can use the same `labelUtil.getLabel()` to get the default label

caritaou: @matthoward366 Do you have any input on this? 
Its not clear where the separation is between canvas and properties with this action. It doesnt make sense to import `labelUtil` in properties and pass that into the action just to get the default label, which is added in the canvas translation file. So maybe my previous comment doesnt make sense..

matthoward366: We don't want to use `labelUtil` since that is canvas specific.  I would think we'd want to pass in the label into the action.  Then we wouldn't pass `propertiesActionLabelHandler, intl` in.  We'd call the callback when the action is added to the stack.  If we do it here the consumer app wouldn't really know which label to get.

matthoward366 (approved):
Looks good.

Issue #902, opened 2022-01-06
Allow for resizable lower panel


nabilIbm:
DSAG approved this pattern, still in the process to get this published.
 https://media.github.ibm.com/user/53906/files/5b5d6f80-275b-11ec-8ea6-9d0e5956eb2b

PR #907, opened 2022-01-07
#902 Making bottom panel resizable

nabilIbm:
Fixes: #902 

           in the file; or

           it.


tomlyn (changes requested):
When I try this the panel jumps from where he mouse pointer position is localed
Bottom Panel

tomlyn (changes requested):

canvas_modules/harness/cypress/support/index.js
```diff
@@ -28,6 +28,7 @@ import "./canvas/toolbar-cmds";
 import "./canvas/utils-cmds";
 import "./canvas/verification-cmds";
 import "./canvas/operation-cmds";
+import "../integration/canvas/bottom-panel";
```
tomlyn, line 31: Why is this here? Do we need it? We don't list any of the other 'integration' files here which ought to be a clue that bottom-panel is not needed here.  This seems to be causing the testcase to run for a second time. Are you not seeing that when you run the bottom-panel test in Cypress?

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
```diff
@@ -746,6 +746,12 @@ Cypress.Commands.add("verifyNodeDimensions", (nodeId, width, height) => {
 		});
 });
 
+Cypress.Commands.add("verifyBoottomPanelHeight", (height) => {
```
tomlyn, line 749: Typo verifyBoottomPanelHeight should be verifyBottomPanelHeight.

canvas_modules/harness/cypress/support/canvas/operation-cmds.js
```diff
@@ -38,3 +38,15 @@ Cypress.Commands.add("selectWithRegion", (x1, y1, x2, y2) => {
 	});
 
 });
+
+Cypress.Commands.add("moveBottomPanelDivider", (yOffset) => {
+// movementY needs to be passed in , because within cc-bottom-panel.jsx onMouseMoveY() is checking
+// on mouse movementY amount which is not passed by Cypress.
```
tomlyn, line 44: These comments should be indented the same amount as the code below it.

canvas_modules/harness/cypress/integration/canvas/bottom-panel.js
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+describe("Testing bottom panel", function() {
+	beforeEach(() => {
+		cy.visit("/");
+		cy.openCanvasDefinition("commentColorCanvas.json");
+	});
+
+	it("Testing bottom pannel seizing", function() {
```
tomlyn, line 22: "pannel seizing"  should be "panel sizing"

canvas_modules/harness/cypress/integration/canvas/bottom-panel.js
```diff
@@ -0,0 +1,33 @@
...
+	// Open right side menu & toggle bottomPanel
```
tomlyn, line 23: This comment should be indented the same amount as the code below it. Also, the comment is wrong because the code is not opening the right side menu and toggling bottom panel. You can probably remove this comment.

tomlyn (changes requested):

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
```diff
@@ -199,5 +183,44 @@ $bottom-panel-height: 393px;
...
+	}
+}
 
+.bottom-panel-contents {
+	background-color: #FFFFFF;
+	flex: 1;
+	width: 1510px;
+	margin: auto;
+	user-select: none;
+}
```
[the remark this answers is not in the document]
matthoward366: I don't know why `background-color: #FFFFFF;` is set.  This should be up to the host application to set.  Why is `width` hardcoded?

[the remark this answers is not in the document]
tomlyn: Good catch @matthoward366   We set the default background-color of the palette and right-flyout panel to `$ui-01` , which the host app can override if it wishes, so I suggest you, Nabil, change it to that.

As for the width, that is working OK in the current code in master where the width of the bottom panel will automatically adjust depending on the size of the palette and size of the right-side flyout (if present). We need to make it so that behavior remains.

tomlyn (changes requested):
Can you do the following please:
1. Remove the code that tries to position the graphic in the center of the canvas area
2. Change the text case in context-menu.js back to the way it was before and change this test case so it is skipped -- that is, instead of saying
```
it("Test context menu position is moved correctly with bottom panl open", function() {
```
it should say 
```
it.skip("Test context menu position is moved correctly with bottom panl open", function() {
```
3. In the text for that test case change `panl` to `panel`
4. The comments in operation-cmds.js need to be indented. I mentiond that in my previous review comment but you seem to have ignored it.

canvas_modules/harness/cypress/integration/canvas/context-menu.js
```diff
@@ -88,7 +88,7 @@ describe("Test of context menu", function() {
 		cy.verifyNumberOfSelectedObjects(0);
 	});
 
-	it("Test context menu position is moved correctly with bottom panl open", function() {
+	it.skip("Test context menu position is moved correctly with bottom panl open", function() {
```
tomlyn, line 91: Please add a comment here to indicate why we are skipping this test. Say something like: 
"This test is being skipped because the canvas area needs to be fixed. At the moment it stretches underneath the bottom panel to the bottom of the common canvas div. It needs to be fixed so that, when the bottom panel is open, the canvas area only extends down to where the bottom panel begins."

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
```diff
@@ -88,15 +83,34 @@ $bottom-panel-height: 393px;
...
+		background: $interactive-01;
+	}
+}
+
+.bottom-panel-contents {
+	background-color: $ui-01;
+	width: 100%;
+	user-select: none;
+	height: 100%;
+
```
tomlyn, line 111: Please add a min-height property of 75px.  (I'm not sure if this needs to be in this rule set or the .bottom-panel rule set so please test it out to make sure your new property works OK. It should  restricts the user from dragging the bottom panel to be less than 75px high)

tomlyn (changes requested):

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
tomlyn: Are you setting "min-height" ? In CSS "min-height" (which means minimum height) is different to "height".  You need to set it on `.bottom-panel` for it to take effect.

canvas_modules/harness/cypress/integration/canvas/context-menu.js
tomlyn: You seem to have resolved this review comment without adding the comment I suggested to the code.

tomlyn (changes requested):

tomlyn (approved):

Issue #988, opened 2022-04-01
Korean Translations 


sara-akhtar:
Adding a new language - Korean

tomlyn:
Implemented in 12.11.0

PR #989, opened 2022-04-01
#988 Korean Translations

sara-akhtar:


matthoward366 (changes requested):
You'll need to also update the index.js files in each folder to enable the language.

matthoward366 (approved):

Issue #1044, opened 2022-05-04
Draggable border in script panel is too thick


nabilIbm:
Ref : 3334

The draggable border for the bottom panel is too thick. It should be reduced to 1px while keeping the ability for the user to easily hover over the border (to be able to size the panel).

PR #1045, opened 2022-05-04
#1044 Draggable border in script panel is too thick

nabilIbm:

           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
+++ b/canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
@@ -95,9 +95,10 @@ $bottom-panel-height: 393px;
 }
 
 .bottom-panel-drag {
-	background: $palette-border-color;
+	border-top: $palette-border-color;
 	cursor: row-resize;
 	flex: 0 0 2px;
+	border-top-width: 1px;
 	transition: all 0.2s ease-in;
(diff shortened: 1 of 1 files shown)


matthoward366:
Screen Shot 2022-05-04 at 7 39 33 AM

There seems like something wrong with the borders yet.  There is a gap between the bottom-panel-drag and top border of bottom-panel.

nabilIbm:
@matthoward366 Can you please check
default case :
image
Hover case :
image

matthoward366:
Screen Shot 2022-05-04 at 10 25 30 AM

The border is still too thick.  You setting it to white which is hiding how thick it actually is.

tomlyn:
@matthoward366 Regarding the border width. If the div is sized to be only one pixel width it makes the "hover over" area quite small which makes it difficult for the user to get the mouse pointer to change to the sizing cursor. So making part of the div width that is colored like the panel seems reasonable - no?

matthoward366:
We don't know the color of the bottom panel since consumers can make the content whatever they want.  In some cases it might not be white which would look odd.

matthoward366:
I think we are good on the using the same color for the background and leaving the draggable div.  We just need to update it to use the token instead.

matthoward366 (approved):
lgtm

PR #1151, opened 2022-08-10
#1150 Show warning in selectcolumn control when invalid value is selected

### LINKED ISSUE #1150: Show warning in selectcolumn control when invalid value is selected
Ref: 12083
Currently common-properties is filtering out invalid enum items and doesn't display them in the UI.
Here is what needs to change

- Still continue showing the invalid enum item
- Show a warning that the value isn't value. This could be specific to the control or something we add to the control messages.
- When the user tries to select another value that enum item shouldn't show up in the list.

For example - `age` doesn't exist in dropdown options. In this case, show selected value as "age" but display warning.
Screen Shot 2022-08-10 at 11 02 09 AM

nmgokhale:
Fixes #1150 
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/properties-controller.js
+++ b/canvas_modules/common-canvas/src/common-properties/properties-controller.js
@@ -49,7 +49,6 @@ export default class PropertiesController {
 		this.visibleDefinitions = {};
 		this.enabledDefinitions = {};
 		this.validationDefinitions = {};
-		this.colDoesExistsDefinitions = {};
 		this.requiredDefinitionsIds = [];
 		this.filterDefinitions = {};
 		this.filteredEnumDefinitions = {};
@@ -154,7 +153,7 @@ export default class PropertiesController {
 			this.saveControls(controls); // saves controls without the subcontrols
 			this._parseSummaryControls(controls);
 			this.parsePanelTree();
-			conditionsUtil.injectDefaultValidations(this.controls, this.validationDefinitions, this.colDoesExistsDefinitions, this.requiredDefinitionsIds, intl);
+			conditionsUtil.injectDefaultValidations(this.controls, this.validationDefinitions, this.requiredDefinitionsIds, intl);
 			let datasetMetadata;
 			let propertyValues = {};
 			if (this.form.data) {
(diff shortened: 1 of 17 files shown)


canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -771,19 +771,21 @@ export default class PropertiesController {
 	*
 	* @param {object} propertyId. required.
 	* @param {boolean} showErrors. optional. Set to false to run conditions without displaying errors in the UI
+	* @param {boolean} onlyValidateInvalidFieldDefinition. optional. Set to true to validate "colDoesExists" condition
 	*    Defaults to true to always display errors
 	*/
-	validateInput(propertyId, showErrors = true) {
-		conditionsUtil.validateInput(propertyId, this, showErrors);
+	validateInput(propertyId, showErrors = true, onlyValidateInvalidFieldDefinition = false) {
```
matthoward366, line 777: Instead of adding a new parameter can we assume of showErrors=false we'd only run the fieldValidation condition?  If we keep the additional parameter we should make it an object instead so we could add more validation if necessary in the future.

nmgokhale: I can validate invalid field definitions when showErrors= false. But then, I have to define a new variable for displaying warning msgs -

```
let showInvalidFieldDefinitionErrors = false; // new variable for displaying warnings even when showErrors=false
if (!showErrors) {
	const invalidFieldDefinition = validations.find((val) => get(val, "definition.validation.id", "").includes("validField_"));
	validations = (typeof invalidFieldDefinition === "undefined") ? [] : [invalidFieldDefinition];
	showInvalidFieldDefinitionErrors = true;
}


// Updated this condition to show warnings
if (!showErrors && !showInvalidFieldDefinitionErrors) {
	errorMessage.displayError = false;
}
```

nmgokhale: This approach may not work because I am replacing all validations with "colDoesExists" validations when `showErrors=false`. So it ignores other validations which regressed this - 
Save button should be disabled because required field is empty. But this condition was never checked. 
Screen Shot 2022-08-18 at 12 11 51 PM

matthoward366 (approved):
looks good

Issue #1235, opened 2022-11-02
Multiselect edit tables unable to scroll to bottom rows


JesusGuerrero:
ref 3401

Recreate:
1. In the test harness, modify the structuretable parameterDef to limit to 5 rows
2. Click add columns and select all fields
3. Once the multiselect header appears, the bottom rows are not visible

<img width="666" alt="image" src="https://media.github.ibm.com/user/11632/files/9797c180-4316-11ed-936d-b871cefd3cc4">

PR #1236, opened 2022-11-02
#1235 Multiselect bug; unable to see rows at the bottom

JesusGuerrero:

           in the file; or

           it.


canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
```diff
@@ -530,6 +531,7 @@ VirtualizedTable.defaultProps = {
 
 VirtualizedTable.propTypes = {
 	tableLabel: PropTypes.string,
+	tableHeight: PropTypes.number,
```
caritaou, line 534: This should be a required prop

canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
```diff
@@ -465,6 +465,7 @@ class VirtualizedTable extends React.Component {
 		// It does a direct DOM manipulation to the parent, outside React's VirtualDOM.
 		// Since the actual DOM is not available when unit testing, we are passing in a default
 		// width of 500 and a default height of 300.
+		const tableHeight = this.props.tableHeight || defaultTestHeight;
```
caritaou, line 468: With these changes, do we still need a `defaultTestHeight`?

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -514,6 +515,10 @@ class FlexibleTable extends React.Component {
 			)
 			: null;
 
+		var tableHeight = this.props.data.length * ROW_HEIGHT;
```
caritaou, line 518: Can you change this to `let` instead?

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -249,7 +250,7 @@ class FlexibleTable extends React.Component {
 		if (Array.isArray(this.props.data) && this.props.data.length < rows) {
 			newHeight = (rowHeight * this.props.data.length + headerHeight) + "rem";
 		} else if (rows > 0) {
-			newHeight = (rowHeight * rows + headerHeight) + "rem";
+			newHeight = (rowHeight * rows + headerHeight + (this.props.selectedEditRow ? REMOF2ROWS : 0)) + "rem";
```
caritaou, line 253: Instead of `REMOF2ROWS`, can you use the `rowHeight` and `headerHeight` defined here to indicate that we need to add a header and row to accommodate the multiselect table? Maybe a new const like:
```
const multiSelectTableHeight = rowHeight + headerHeight;
```
for readability

caritaou (changes requested):
The main concern with these changes is how it affects table that have `noAutoSize` set to true. 
```
if (!this.props.noAutoSize) {
	heightStyle = { height: this.state.tableHeight };
}
```

In Common Properties, this is only used within field-picker.jsx. You can see how the tables are affected in the structureTable paramDef 
image

Since FlexibleTable is exported as its own components, there could be other applications using this as well.

canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
JesusGuerrero: i think yes, the old logic `height ? height : defaultTestHeight` would work off of the Table calculating the height, but now we calculate the height and data.length can be 0 so it will force a height 0 which may affect testing. This is not something I'm 100% sure on, this would leave it in

caritaou: > but now we calculate the height and data.length can be 0 so it will force a height 0

Since that is what the code does then the test should also behave the same. The `defaultTestHeight` was introduced because AutoSizer requires the DOM to exist to fit itself into and will return a `height` to set in the style. But because jest uses the virtual DOM, there was no parent container to fit itself into. Now that a `height` is being calculated, we won't ever need to read the `height` from AutoSizer, which in turn makes me believe its safe to remove the `defaultTestHeight` but will need to run the tests to make sure

JesusGuerrero: ok, I'm comfortable to proceed with removal if you are aligned. Will test additional ways.

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -514,6 +519,15 @@ class FlexibleTable extends React.Component {
 			)
 			: null;
 
+		let tableHeight = 0;
+		if (this.state.rows !== -1 && this.state.tableHeight) {
+			const remHeight = Number(this.state.tableHeight.match(/^[0-9]{1,2}/i)[0]);
```
caritaou, line 524: Are you able to use `parseInt()` instead?

nmgokhale:
Set `rows: 5` in paramDef.
When there are 5+ rows in the table, and selecting multiple rows, I see total 5 rows in the table - 
Screen Shot 2022-11-16 at 3 51 08 PM

But if there are less than 5 rows in the table - 
Screen Shot 2022-11-16 at 3 51 44 PM

And I select multiple rows, it shows only 2 rows and I have to scroll down to see remaining rows -
Screen Shot 2022-11-16 at 3 53 07 PM

JesusGuerrero:
Thanks @nmgokhale ... fixed that bug, not it shows correctly

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -514,6 +517,19 @@ class FlexibleTable extends React.Component {
...
+		const multiSelectEditRowsPixels = multiSelectEditRowsRem * ONE_REM_HEIGHT;
+		if (this.state.rows !== -1 && this.state.tableHeight) {
+			const remHeight = parseInt(this.state.tableHeight.match(/^[0-9]{1,2}/i)[0], 10);
+			tableHeight = (remHeight - (this.props.selectedEditRow ? multiSelectEditRowsRem : 0)) * ONE_REM_HEIGHT;
+		} else if (this.state.rows === -1) {
+			if (this.state.dynamicHeight && this.state.dynamicHeight !== -1) {
+				tableHeight = this.state.dynamicHeight - (this.props.selectedEditRow ? multiSelectEditRowsPixels : 0);
+			} // else how do we handle this.state.tableHeight = "100vh"?
+		}
+		console.log(tableHeight);
```
nmgokhale, line 531: Please remove this console.log statement

JesusGuerrero: 😂 ... happens when you push code in middle of the night

nmgokhale (approved):

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -514,6 +517,18 @@ class FlexibleTable extends React.Component {
 			)
 			: null;
 
+		let tableHeight = 0;
+		const multiSelectEditRowsRem = 2 * REM_HEADER_HEIGHT; // multi-select adds two rows when selectedEditRow
+		const multiSelectEditRowsPixels = multiSelectEditRowsRem * ONE_REM_HEIGHT;
+		if (this.state.rows !== -1 && this.state.tableHeight) {
+			const remHeight = parseInt(this.state.tableHeight.match(/^[0-9]{1,2}/i)[0], 10);
```
caritaou, line 524: I think you can remove match and just have `parseInt(this.state.tableHeight, 10);`

JesusGuerrero: yeesss!

nmgokhale (approved):

canvas_modules/common-canvas/src/common-properties/components/flexible-table/flexible-table.jsx
```diff
@@ -514,6 +517,18 @@ class FlexibleTable extends React.Component {
...
+		let tableHeight = 0;
+		const multiSelectEditRowsRem = 2 * REM_HEADER_HEIGHT; // multi-select adds two rows when selectedEditRow
+		const multiSelectEditRowsPixels = multiSelectEditRowsRem * ONE_REM_HEIGHT;
+		if (this.state.rows !== -1 && this.state.tableHeight) {
+			const remHeight = parseInt(this.state.tableHeight, 10);
+			tableHeight = (remHeight - (this.props.selectedEditRow ? multiSelectEditRowsRem : 0)) * ONE_REM_HEIGHT;
+		} else if (this.state.rows === -1) {
+			if (this.state.dynamicHeight && this.state.dynamicHeight !== -1) {
+				tableHeight = this.state.dynamicHeight - (this.props.selectedEditRow ? multiSelectEditRowsPixels : 0);
+			} // else how do we handle this.state.tableHeight = "100vh"?
```
caritaou, line 529: Does this comment still need to be addressed? If not, the else if from line 526 and 527 can be combined

JesusGuerrero: I didn't find a test case for the "100vh", it can be consolidated.

caritaou (approved):

nmgokhale (approved):

PR #1343, opened 2023-01-27
#1342 config convert values

### LINKED ISSUE #1342: Add new properties config option to convert the currentParameter values data type to what is defined in the parameterDef
ref 3471

There may be instances where a boolean parameter gets initiated with `"true"` instead of `true`. If the new config option is enabled, we will convert the value from a string to boolean when reading the values.

caritaou:
closes #1342 
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/__tests__/common-properties/utils/property-utils-test.js
+++ b/canvas_modules/common-canvas/__tests__/common-properties/utils/property-utils-test.js
@@ -15,12 +15,15 @@
  */
 
 import { expect } from "chai";
+import { propertyOf } from "lodash";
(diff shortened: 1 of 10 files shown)


matthoward366 (commented):
Shouldn't this be false?  

Screenshot 2023-01-27 at 6 58 07 AM

When I save the values back out shouldn't they have the converted values?  I'm seeing the original values.

caritaou:
Did you have this option `convertValueDataTypes` set to true? I had that set to false by default to be align with how it works today
image

matthoward366 (approved):
Looks good.  I forgot the toggle.

Issue #1387, opened 2023-03-04
Update Tooltips in textFields


nabilIbm:
ref : 3201

<img width="478" alt="image" src="https://user-images.githubusercontent.com/89609176/222857808-b7fcc8c2-3b33-499d-b37c-99a64dc503a2.png">

nmgokhale:
Fixed in PR https://github.com/elyra-ai/canvas/pull/1409

PR #1388, opened 2023-03-04
#1387 Update Tooltips in textFields

nabilIbm:

           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/components/truncated-content-tooltip/truncated-content-tooltip.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/components/truncated-content-tooltip/truncated-content-tooltip.jsx
@@ -24,6 +24,8 @@ import { has } from "lodash";
 export default class TruncatedContentTooltip extends React.Component {
 
 	render() {
+		const input = this.triggerRef && this.triggerRef.getElementsByTagName("input")[0];
+		const truncatedInput = !(input && input.scrollWidth > input.clientWidth);
 		let tooltipText = this.props.tooltipText;
 		if (typeof this.props.tooltipText !== "object") {
 			tooltipText = String(this.props.tooltipText);
@@ -35,16 +37,20 @@ export default class TruncatedContentTooltip extends React.Component {
 		);
 		return (
 			<div className="properties-truncated-tooltip">
-				<Tooltip
-					id={`${uuid4()}-properties`}
-					tip={tooltip}
-					direction="bottom"
-					className="properties-tooltips"
-					disable={has(this.props, "disabled") ? this.props.disabled : true}
-					showToolTipIfTruncated
-				>
-					{this.props.content}
-				</Tooltip>
+				{truncatedInput
+					?	<Tooltip
+						id={`${uuid4()}-properties`}
+						tip={tooltip}
+						direction="bottom"
+						className="properties-tooltips"
+						disable={has(this.props, "disabled") ? this.props.disabled : true}
+						showToolTipIfTruncated
+					>
+						{this.props.content}
+					</Tooltip>
+					: <>
+						{this.props.content}
+					</>}
 			</div>
 		);
 	}
--- a/canvas_modules/common-canvas/src/common-properties/components/truncated-content-tooltip/truncated-content-tooltip.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/components/truncated-content-tooltip/truncated-content-tooltip.jsx
@@ -22,10 +22,27 @@ import { has } from "lodash";
 
 // Reusable component to show tooltip if the content is truncated
 export default class TruncatedContentTooltip extends React.Component {
+	// Return true if the string can be displayed in the available space
+	// Return false if the string is truncated and ellipsis is shown on the UI
+	// (offsetWidth) is a measurement in pixels of the element's CSS width, including any borders, padding, and vertical scrollbars
+	// (scrollWidth) value is equal to the minimum width the element would require
+	//  in order to fit all the content in the viewport without using a horizontal scrollbar
+	canDisplayFullText(elem) {
+		if (elem) {
+			const firstChildWidth = elem.firstChild && elem.firstChild.scrollWidth ? elem.firstChild.scrollWidth : 0;
+			const displayWidth = elem.offsetWidth;
+			let fullWidth = firstChildWidth;
+			if (firstChildWidth === 0) {
+				fullWidth = elem.scrollWidth;
+			}
+			const canDisplayFullText = fullWidth <= displayWidth;
+			return canDisplayFullText;
+		}
+		return false; // Show tooltip if we cannot read the width (Canvas objects)
+	}
 
 	render() {
-		const input = this.triggerRef && this.triggerRef.getElementsByTagName("input")[0];
-		const truncatedInput = !(input && input.scrollWidth > input.clientWidth);
+		const canDisplayFullText = this.canDisplayFullText(this.props.truncatElem);
 		let tooltipText = this.props.tooltipText;
 		if (typeof this.props.tooltipText !== "object") {
 			tooltipText = String(this.props.tooltipText);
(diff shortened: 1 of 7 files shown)


canvas_modules/common-canvas/src/common-properties/controls/textfield/textfield.jsx
```diff
@@ -109,25 +110,26 @@ class TextfieldControl extends React.Component {
...
+			let content = textInput;
 			let disabled = true;
 			if (value && this.props.state !== STATES.DISABLED) {
 				disabled = false;
 			}
-			const tooltip = (
-				<div className="properties-tooltips">
-					{String(value)}
-				</div>
+			if (this.props.control.icon) {
```
matthoward366, line 118: Why add the icon?  This shouldn't be added for textfields.

matthoward366 (changes requested):
I don't see tooltips working anymore for textfields in tables.

canvas_modules/common-canvas/src/common-properties/controls/textfield/textfield.jsx
```diff
@@ -109,25 +110,26 @@ class TextfieldControl extends React.Component {
...
+						<div className={"properties-field-type-icon"}>
+							<Icon type={this.props.control.icon} />
+						</div>
+						{textInput}
+					</div>);
+			}
+			display = (
+				<TruncatedContentTooltip
+					content={content}
+					disabled={disabled}
```
caritaou, line 130: You're missing the `tooltipText` that is marked as required in this component
image

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -107,7 +107,7 @@ class ToolTip extends React.Component {
 		const canDisplayFullText = this.canDisplayFullText(this.triggerRef);
 		const showToolTip = (
 			// show tooltip if not disabled and showToolTipIfTruncated is false
-			(!this.props.disable && !this.props.showToolTipIfTruncated) ||
+			(!this.props.disable && !this.props.showToolTipIfTruncated && !canDisplayFullText) ||
```
matthoward366, line 110: This is the same as line 112. This doesn't seem right.  Why was this changed?

nabilIbm: I added  it to have the expected label's behavior the same as textfiled (otherwise tooltip will show in ALL header values )
I'll  update the function below then `canDisplayFullText` , see  what needs to be changed to handle both (label & textfiled)

matthoward366: For the table header it might be caused by using `Tooltip` instead of TruncatedContentTooltip

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -119,7 +119,8 @@ class ToolTip extends React.Component {
 	// (scrollWidth) value is equal to the minimum width the element would require
 	//  in order to fit all the content in the viewport without using a horizontal scrollbar
 	canDisplayFullText(elem) {
-		if (elem) {
+		const input = elem.getElementsByTagName("input")[0];
```
matthoward366, line 122: This doesn't seem like the right place for this.  It's too specific to `input` elements and the tooltip component is more generic.  I would probably pass in a new property into the tooltip component to check.  Something like `truncatedElement`.  Then pass the `input` element into that.  It would be optional and fall back to this.props.children if not set.

matthoward366 (changes requested):

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
nabilIbm: I tried it in first but i got all textfield gone from the UI
image

nabilIbm: What i was trying to say in here,  is im using `TruncatedContentTooltip` not `Tooltip` (if this was your comment) and yet labels didn't work 
image

matthoward366: You should update your PR with the correct code then figure out why it doesn't work correctly.

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -30,6 +30,7 @@ class ToolTip extends React.Component {
 		};
 
 		this.pendingTooltip = null;
+		this.truncatedElement = null;
```
matthoward366, line 33: This should be a property passed in by the TextField control.

matthoward366: Then that would be used instead of `elem` in canDisplayFullText.

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -331,6 +332,8 @@ class ToolTip extends React.Component {
 	}
 
 	render() {
+		const inputTruncated = this.triggerRef && this.triggerRef.getElementsByTagName("input")[0];
+		this.truncatedElement = !(inputTruncated && inputTruncated.scrollWidth > inputTruncated.clientWidth);
```
caritaou, line 336: Hi @nabilIbm, this code right here is what's calculating if the Textfield should be truncated or not. This logic should be moved to `TruncatedContentTooltip` instead. Then no changes are required in `Tooltip`. Inside `TruncatedContentTooltip`, if you determine that the `content` needs to be truncated, you will return the `<Tooltip>`, otherwise, just return the `content`

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -440,14 +420,16 @@ ToolTip.propTypes = {
 	disable: PropTypes.bool, // Tooltip will not show if disabled
 	showToolTipIfTruncated: PropTypes.bool, // Set to true to only display tooltip if full text does not fit in displayable width
 	delay: PropTypes.number,
-	showToolTipOnClick: PropTypes.bool
+	showToolTipOnClick: PropTypes.bool,
+	truncatElem: PropTypes.object,
```
matthoward366, line 424: This shouldn't be needed since this is handled in TruncatedTooltip.

canvas_modules/common-canvas/src/common-properties/components/truncated-content-tooltip/truncated-content-tooltip.jsx
```diff
@@ -22,8 +22,27 @@ import { has } from "lodash";
 
 // Reusable component to show tooltip if the content is truncated
 export default class TruncatedContentTooltip extends React.Component {
+	// Return true if the string can be displayed in the available space
+	// Return false if the string is truncated and ellipsis is shown on the UI
+	// (offsetWidth) is a measurement in pixels of the element's CSS width, including any borders, padding, and vertical scrollbars
+	// (scrollWidth) value is equal to the minimum width the element would require
+	//  in order to fit all the content in the viewport without using a horizontal scrollbar
+	canDisplayFullText(elem) {
+		if (elem) {
```
matthoward366, line 31: I would expect this to be `elem` or `truncatElem`.  Then textfield passes in `truncatElem`

matthoward366 (changes requested):

caritaou:
Hey @matthoward366, with the changes from this PR, should we also export `TruncatedContentTooltip` in https://github.com/elyra-ai/canvas/blob/main/canvas_modules/common-canvas/src/index.js? Some applications using `Tooltip` pass in `showToolTipIfTruncated`, so they will need to use `TruncatedContentTooltip` instead

matthoward366:
@caritaou Forgot about that.  Why even have `TruncatedContentTooltip`?  It probably makes more sense to adjust Tooltip to deal with the css needed from `TruncatedContentTooltip`.

nabilIbm:
Do we just keep `Tooltip` in both textfiled (text field)  & virtualized-table (for labels) and adjust them in css to inherit `TruncatedContentTooltip` behavior ?

matthoward366:
You'd `TruncatedContentTooltip` completely and correct by updating Tooltip to use the css from TruncatedContentTooltip when `showToolTipIfTruncated` is passed in.

matthoward366:
Replaced by https://github.com/elyra-ai/canvas/pull/1409

Issue #1402, opened 2023-03-17
Accessibility: Tooltip element's role 'none' is not a widget role


nmgokhale:
Ref: 3485

Fix this violation in all common-properties tooltips.

<img width="667" alt="tooltip-violation" src="https://user-images.githubusercontent.com/25124000/225778592-1f9f3bd1-3843-4adf-afa3-7d19060a50d2.png">

![Screenshot 2023-03-16 at 5 04 14 PM](https://user-images.githubusercontent.com/25124000/225778698-1dc02423-7b70-480c-aac6-62ad27c26b1f.png)

PR #1403, opened 2023-03-17
#1402 Accessibility: Fixed Tooltip element's role 'none' is not a widget role

nmgokhale:
Fixes: #1402 

Followed same pattern as [carbon tooltip](https://v10.carbondesignsystem.com/components/tooltip/code/). 
- For toolbar - Added `aria-labelledby` attribute to the `tooltip-trigger` element.
- For common-properties tooltips - Added `aria-labelledby, aria-expanded, aria-controls` attributes to the `tooltip-trigger` element. 

Follwing [ARIA example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#examples) -
Added `role="tooltip", id` attributes to actual tooltip content having `className="common-canvas-tooltip"`.

**Before -**
Screenshot 2023-03-16 at 5 04 14 PM

**After -**
Screenshot 2023-03-16 at 5 17 11 PM

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
@@ -24,7 +24,6 @@ import { injectIntl } from "react-intl";
 import defaultMessages from "../../../../locales/common-properties/locales/en.json";
 
 import { isEmpty, differenceBy, mapValues } from "lodash";
-import { v4 as uuid4 } from "uuid";
 import classNames from "classnames";
 
 import PropTypes from "prop-types";
@@ -53,10 +52,6 @@ class VirtualizedTable extends React.Component {
 			columns: this.props.columns,
 			columnResized: false
 		};
-		this.uuidForColumns = this.state.columns.reduce(function(prev, column) {
-			prev[column.key] = uuid4();
-			return prev;
-		}, {});
 		this.virtualizedTableRef = React.createRef();
 		this.lastChecked = isEmpty(props.rowsSelected) ? null : props.rowsSelected.slice(-1).pop();
 
@@ -249,7 +244,7 @@ class VirtualizedTable extends React.Component {
 			? null
 			: (<div className="properties-vt-info-icon-tip">
 				<Tooltip
-					id={`properties-tooltip-${this.uuidForColumns[dataKey]}-info`}
+					id="properties-tooltip-info"
 					tip={columnData.description}
 					direction="bottom"
 					className="properties-tooltips"
@@ -259,7 +254,7 @@ class VirtualizedTable extends React.Component {
 				</Tooltip>
 			</div>);
 
-		const tooltipId = `${this.uuidForColumns[dataKey]}-tooltip-column-${dataKey}`;
+		const tooltipId = `tooltip-column-${dataKey}`;
 
 		const resizeElem = columnData.resizable && !this.isLastColumn(dataKey)
 			? (<Draggable
--- a/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
@@ -245,7 +245,7 @@ class VirtualizedTable extends React.Component {
 			? null
 			: (<div className="properties-vt-info-icon-tip">
 				<Tooltip
-					id={`properties-tooltip-${columnData.headerLabel}-info`}
+					id={`properties-tooltip-${columnData.headerLabel.split(" ").join("")}-info`}
 					tip={columnData.description}
 					direction="bottom"
 					className="properties-tooltips"
--- a/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
@@ -53,6 +53,10 @@ class VirtualizedTable extends React.Component {
 			columns: this.props.columns,
 			columnResized: false
 		};
+		this.uuidForColumns = this.state.columns.reduce(function(prev, column) {
+			prev[column.key] = uuid4();
+			return prev;
+		}, {});
 		this.virtualizedTableRef = React.createRef();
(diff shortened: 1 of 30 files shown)


canvas_modules/common-canvas/src/common-properties/controls/textarea/textarea.jsx
```diff
@@ -98,7 +98,7 @@ class TextareaControl extends React.Component {
 
 		let display = textArea;
 		if (this.props.tableControl) {
-			const tooltipId = uuid4() + "-tooltip-column-" + this.props.propertyId.toString();
+			const tooltipId = uuid4() + "-tooltip-column-" + getPropertyIdOfTableControlString(this.props.propertyId);
```
nmgokhale, line 101: Before there was `[Object Object]` in tooltipId  -
Screenshot 2023-03-16 at 5 28 31 PM

After -
Screenshot 2023-03-16 at 5 33 56 PM

canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
```diff
@@ -245,7 +245,7 @@ class VirtualizedTable extends React.Component {
 			? null
 			: (<div className="properties-vt-info-icon-tip">
 				<Tooltip
-					id={`properties-tooltip-${columnData.headerLabel}-info`}
+					id={`properties-tooltip-${columnData.headerLabel.split(" ").join("")}-info`}
```
caritaou, line 248: This is the same as `string.replace(" ", "")`?

canvas_modules/common-canvas/src/common-properties/util/property-utils.js
```diff
@@ -573,6 +573,18 @@ function convertValueDataTypes(currentParameters, paramDefs) {
 	return convertedCurrentParameters;
 }
 
+// Return string value for {name: string, row: integer, col: integer}
+function getPropertyIdOfTableControlString(propertyId) {
```
caritaou, line 577: Note that we do support nesting in properties, for example in structuretable_paramDef:
image

Take a look at `propertiesController.convertNestedPropertyId()`

canvas_modules/common-canvas/src/common-properties/components/virtualized-table/virtualized-table.jsx
matthoward366: This doesn't seem like a good way of creating ids.  There could be other chars in the header label that cause issues and there is no guarantee these are unique in the DOM.  Can we generate a guid for each header instead?  We wouldn't want to generate it on each rerender though.

canvas_modules/common-canvas/src/common-properties/controls/textarea/textarea.jsx
matthoward366: Are we getting new uuid each time we rerender?  That doesn't seem like a good idea.  Should we move the uuid to the constructor so it's only generated once for this class?

canvas_modules/common-canvas/src/common-properties/controls/textfield/textfield.jsx
```diff
@@ -109,7 +109,7 @@ class TextfieldControl extends React.Component {
 
 		let display = textInput;
 		if (this.props.tableControl) {
-			const tooltipId = uuid4() + "-tooltip-column-" + this.props.propertyId.toString();
+			const tooltipId = uuid4() + "-tooltip-column-" + getPropertyIdOfTableControlString(this.props.propertyId);
```
matthoward366, line 112: Same comment as above

canvas_modules/common-canvas/src/common-properties/util/property-utils.js
matthoward366: Take a look at https://github.com/elyra-ai/canvas/blob/main/canvas_modules/common-canvas/src/common-properties/util/control-utils.js#L25.  I think we already have a method that does this.

As long as these id's are unique I don't think we really need to care about them being nested.

canvas_modules/common-canvas/src/common-properties/controls/textarea/textarea.jsx
nmgokhale: Yes there's a new `uuid` for each rerender and all table cells have a different `uuid`. I will generate `uuid` once in constructor and use it for all tooltipIds.

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -406,7 +411,14 @@ class ToolTip extends React.Component {
 		if (tooltipContent || link) {
 			tooltip = (
 				<Portal>
-					<div data-id={this.props.id} className={tipClass} aria-hidden={!this.state.isTooltipVisible} direction={this.props.direction}>
+					<div
+						role="tooltip"
+						id={`${this.uuid}-${this.props.id}`}
```
nmgokhale, line 416: Generating `UUID` in tooltip. Prefix it to tooltip id. Consuming component can pass a generic `id` in props. UUID makes sure tooltipId is unique for all.

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -368,13 +373,13 @@ class ToolTip extends React.Component {
 
 		if ((typeof this.props.tip) === "string") {
 			tooltipContent = (
-				<span id="tooltipContainer">
+				<span id={`${this.uuid}-tooltipContainer`}>
```
matthoward366, line 376: For the `id` do we actually need to set `id`?  I'm curious if it's actually needed.  This seems like it should be a class or not set.

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
nmgokhale: We need an `id` for `aria-labelledby` attribute to work. 
We can remove `data-id` now that `id` will be unique after using uuid. But many jest tests will break after removing data-id, so I kept both id and data-id.

matthoward366: Ah, ok.

matthoward366 (approved):
Looks good

caritaou (approved):

tomlyn (approved):

PR #1578, opened 2023-09-27
#1575 When parameterDef is dynamically updated, don't set INITIAL_LOAD on pre-existing properties

### LINKED ISSUE #1575: Add an option to refresh parameterDef for common-properties
Ref: 3603

Add a new config option in [propertiesInfo object](https://github.com/elyra-ai/canvas/wiki/3.0-Common-Properties-documentation#step-2--set-the-data).

```
id:  "{id}"	            // Unique parameter definition ID [Optional]
```

`id` will be used to compare whether updated parameterDef is provided or a completely new parameterDef is provided to the common-properties.

In updated parameterDef, don't set `type: initial_load` on pre-existing properties.

nmgokhale:
Fixes #1575 

Added a new optional config option `id` in [propertiesInfo object](https://github.com/elyra-ai/canvas/wiki/3.0-Common-Properties-documentation#step-2--set-the-data).

To test this PR, add following code in test harness App.js - https://github.com/elyra-ai/canvas/blob/main/canvas_modules/harness/src/client/App.js#L1940

```
setTimeout(() => {
			console.log("Updating paramDef after 30 seconds");
			const oldParamDef = this.state.propertiesInfo.parameterDef;
			const newParamDef = cloneDeep(oldParamDef);

			const uuid = uuid4();
			const parameterObj = {
				"id": `checkbox-${uuid}`,
				"type": "boolean",
				"default": true
			};

			const parameterInfoObj = {
				"parameter_ref": `checkbox-${uuid}`,
				"label": {
					"default": `checkbox-${uuid}`
				},
				"description": {
					"default": "new added checkbox with undefined value"
				}
			};
			newParamDef.parameters.push(parameterObj);
			newParamDef.uihints.parameter_info.push(parameterInfoObj);
			newParamDef.uihints.group_info[0].parameter_refs.push(`checkbox-${uuid}`);

			const newPropsInfo = cloneDeep(this.state.propertiesInfo);
			newPropsInfo.parameterDef = newParamDef;
			this.setState({ propertiesInfo: newPropsInfo });

		}, 30000);

```
Also add logs in propertyListener - https://github.com/elyra-ai/canvas/blob/main/canvas_modules/harness/src/client/App.js#L1795

```
propertyListener(data) {
		console.log(data);
}
```

After updating current parameterDef every 5 seconds by adding a new property, notice `INITIAL_LOAD` is added only on the latest property. Not on the pre-existing properties.

https://github.com/elyra-ai/canvas/assets/25124000/8f039fba-650f-4110-924e-7aaa79b5ab6f

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/properties-controller.js
+++ b/canvas_modules/common-canvas/src/common-properties/properties-controller.js
@@ -28,7 +28,7 @@ import { STATES, ACTIONS, CONDITION_TYPE, PANEL_TREE_ROOT, CONDITION_MESSAGE_TYP
 import CommandStack from "../command-stack/command-stack.js";
 import ControlFactory from "./controls/control-factory";
 import { Type, ParamRole, ControlType, ItemType } from "./constants/form-constants";
-import { has, cloneDeep, assign, isEmpty, isEqual, isUndefined, get } from "lodash";
+import { has, cloneDeep, assign, isEmpty, isEqual, isUndefined, get, difference } from "lodash";
 import Form from "./form/Form";
 import { getConditionOps } from "./ui-conditions/condition-ops/condition-ops";
 import { DEFAULT_LOCALE } from "./constants/constants";
@@ -427,7 +427,7 @@ export default class PropertiesController {
 				let differentProperties = [];
 				if (sameParameterDefRendered) {
 					// When a parameterDef is dynamically updated, set difference between old and new controls
-					differentProperties = PropertyUtils.getObjectDifference(this.controls, this.prevControls);
+					differentProperties = difference(Object.keys(this.controls), Object.keys(this.prevControls));
 				}
 
 				if (sameParameterDefRendered && differentProperties.indexOf(control.name) === -1) {
--- a/canvas_modules/common-canvas/src/common-properties/properties-controller.js
+++ b/canvas_modules/common-canvas/src/common-properties/properties-controller.js
@@ -55,7 +55,7 @@ export default class PropertiesController {
 		this.allowChangeDefinitions = {};
 		this.conditionalDefaultDefinitions = {};
 		this.panelTree = {};
-		this.oldControls = {};
+		this.prevControls = {};
 		this.controls = {};
 		this.actions = {};
 		this.customControls = [];
@@ -111,23 +111,15 @@ export default class PropertiesController {
 		return this.editorSize;
 	}
 
-	setOldParameterDefId(oldParameterDefId) {
-		this.oldParameterDefId = oldParameterDefId;
-	}
-
-	getOldParameterDefId() {
-		return this.oldParameterDefId;
-	}
-
 	/**
-	* @param {string} parameterDefId - Optional ID from propertiesInfo
+	* @param {string} id - Optional ID from propertiesInfo
 	*/
-	setParameterDefId(parameterDefId) {
-		this.parameterDefId = parameterDefId;
+	setId(id) {
+		this.id = id;
 	}
 
-	getParameterDefId() {
-		return this.parameterDefId;
+	getId() {
+		return this.id;
 	}
 
 	setLight(light) {
@@ -154,16 +146,6 @@ export default class PropertiesController {
 		return this.propertiesConfig.containerType === "Tearsheet";
 	}
 
-	// Compare old & new controls, set an array of different properties
-	setDifferentProperties() {
-		const differentProperties = PropertyUtils.getObjectDifference(this.controls, this.oldControls);
-		this.differentProperties = differentProperties;
-	}
-
-	getDifferentProperties() {
-		return this.differentProperties;
-	}
-
 	setParamDef(paramDef) {
 		const containerType = get(this.getPropertiesConfig(), "containerType");
 		const convertTypes = get(this.getPropertiesConfig(), "convertTypes");
@@ -180,7 +162,7 @@ export default class PropertiesController {
 		this.reactIntl = intl;
 		// set initial property values
 		if (this.form) {
-			this.oldControls = this.controls;
+			this.prevControls = this.controls;
 			this.controls = {};
 			this.setControlStates({}); // clear state
 			this.setErrorMessages({}); // clear messages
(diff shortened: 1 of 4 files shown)


canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -110,6 +111,25 @@ export default class PropertiesController {
 		return this.editorSize;
 	}
 
+	setOldParameterDefId(oldParameterDefId) {
```
matthoward366, line 114: Instead of having these methods can't we just pass the ID in the setForm method?  We can still save the Id but we shouldn't need to keep the old and newIds stored.  We also shouldn't need these methods since only the controller needs to care about the Ids.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -134,6 +154,16 @@ export default class PropertiesController {
 		return this.propertiesConfig.containerType === "Tearsheet";
 	}
 
+	// Compare old & new controls, set an array of different properties
+	setDifferentProperties() {
+		const differentProperties = PropertyUtils.getObjectDifference(this.controls, this.oldControls);
```
matthoward366, line 159: Would lodash.differenceBy do this same thing?  I think we just need to compare the keys of the objects.  Right?  We only need to compute the once so I don't know if we need to store differences as a class variable.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -150,6 +180,7 @@ export default class PropertiesController {
 		this.reactIntl = intl;
 		// set initial property values
 		if (this.form) {
+			this.oldControls = this.controls;
```
matthoward366, line 183: instead of old maybe `prevControls` or maybe even just the controlIds if that's all we care about.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -58,6 +58,7 @@ class PropertiesMain extends React.Component {
 		this.propertiesController.setLight(props.light);
 		this.propertiesController.setAppData(props.propertiesInfo.appData);
 		this.propertiesController.setExpressionInfo(props.propertiesInfo.expressionInfo);
+		this.propertiesController.setParameterDefId(props.propertiesInfo.parameterDefId);
```
matthoward366, line 61: Should this just be `id`.  Using parameterDefId makes me think it should go inside the paramDef.

canvas_modules/common-canvas/src/common-properties/properties-controller.js
nmgokhale: Tried using `differenceBy(this.controls, this.prevControls, isEqual)` but its returning an empty array.

matthoward366: You'd probably write your own function to compare the control names.

matthoward366: or maybe just pass in the keys since that's really what you to know the difference on.  Then I think you can use just difference

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -107,12 +108,20 @@ class PropertiesMain extends React.Component {
 				(newProps.propertiesInfo.formData && !isEqual(newProps.propertiesInfo.formData, this.props.propertiesInfo.formData)) ||
 				(newProps.propertiesInfo.parameterDef && !isEqual(newProps.propertiesInfo.parameterDef, this.props.propertiesInfo.parameterDef)) ||
 				(newProps.propertiesInfo.appData && !isEqual(newProps.propertiesInfo.appData, this.props.propertiesInfo.appData))) {
+				// Save prevId first before setting the new one from propertiesInfo
+				this.propertiesController.prevId = this.propertiesController.getId();
```
matthoward366, line 112: Why do we need to store the id?  Can we just compare the new vs prev properties like we do other the parameterDef.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -107,12 +108,20 @@ class PropertiesMain extends React.Component {
...
+					this.propertiesController.setId(newProps.propertiesInfo.id);
+				}
 				this.setForm(newProps.propertiesInfo);
 				const newEditorSize = this.propertiesController.getForm().editorSize;
 				if (this.state.editorSize !== newEditorSize) {
 					this.setState({ editorSize: newEditorSize });
 				}
-				this.currentParameters = this.propertiesController.getPropertyValues();
+				// Reset property values for new parameterDef
+				if (has(newProps.propertiesInfo, "id") && (this.propertiesController.prevId !== this.propertiesController.getId())) {
```
matthoward366, line 122: Why use `has`.  Can't we just do `if (newProps.propertiesInfo.id ...)`

canvas_modules/common-canvas/src/common-properties/properties-controller.js
```diff
@@ -409,7 +422,19 @@ export default class PropertiesController {
...
-				this.updatePropertyValue(propertyId, controlValue, true, UPDATE_TYPE.INITIAL_LOAD);
+				// When parameterDef is dynamically updated, don't set INITIAL_LOAD on pre-existing properties
+				const sameParameterDefRendered = this.prevId === this.getId();
+				let differentProperties = [];
+				if (sameParameterDefRendered) {
+					// When a parameterDef is dynamically updated, set difference between old and new controls
+					differentProperties = difference(Object.keys(this.controls), Object.keys(this.prevControls));
+				}
+
+				if (sameParameterDefRendered && differentProperties.indexOf(control.name) === -1) {
```
matthoward366, line 433: Would using `includes` be better here?

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
nmgokhale: We store everything under `propertiesInfo` in `propertiesController` - https://github.com/elyra-ai/canvas/blob/main/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx#L59C3-L60C84

Followed the same pattern for `propertiesInfo.id`.

matthoward366: Right, but we really don't need this for later on like the others.

matthoward366 (approved):
lgtm

PR #1587, opened 2023-10-12
Notification panel scrolling issue

### LINKED ISSUE #135: Notification center unable to scroll on narrow window
When the window is too narrow, the notification center is no longer able to scroll to see later messages

nmgokhale:
Fixes #135 

Increased max-height to cover 100% of the screen height.
When there are 8+ notification messages, notification panel hides right side of bottom panel. Notice the `Refresh` and `Compare` buttons in the bottom panel are hidden behind notification panel.
Screenshot 2023-10-12 at 3 12 22 PM

 
           in the file; or

           it.


caritaou (commented):
notification
I noticed that the height of the buttons are off when the notification is scrolling, and once theres enough space on the screen, then the full height of the buttons are shown. I know its a small thing but are you able to make the full height of the buttons visible even when scrolling the messages?

nmgokhale:
Figured out a way to fix this issue without using 100% screen height. Now the max-height of notification panel is 635px (same as before) and content is scrollable. Also fixed the button height issue. Thanks Carita.


I will update my branch after all other approved PRs are merged in main.

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
```diff
@@ -82,11 +86,11 @@ $toolbar-button-height: 41px; // Allow one extra pixel for the toolbar border
 
 .notification-panel-messages-container {
 	width: inherit;
-	max-height: calc(100% - #{$notification-panel-header-height});
-	height: fit-content;
+	max-height: calc(100% - #{$notification-panel-header-height} - 10px); // Extra 10px makes sure full height of notification-panel-button-container is visible in narrow window
```
tomlyn, line 89: Is the 10px here just an arbitrary number? If it is the button height can we create a variable set to 10px with an appropriate name so the calculation expression is readable?  

Also, although the word 'narrow' is used in the issue title for this work it is the wrong word since narrow refers to a a small , or constrained, width of something. I don't think there is an English equivalent of narrow in the vertical direction but maybe 'short' would be a better word.

Thirdly, did you experiment with `vh` values for this instead of a calculation using pixels -- as mentioned in my comment in the issue?

tomlyn (changes requested):

canvas_modules/common-canvas/src/notification-panel/notification-panel.scss
nmgokhale: > Is the 10px here just an arbitrary number? If it is the button height can we create a variable set to 10px with an appropriate name so the calculation expression is readable?

> Also, although the word 'narrow' is used in the issue title for this work it is the wrong word since narrow refers to a a small , or constrained, width of something. I don't think there is an English equivalent of narrow in the vertical direction but maybe 'short' would be a better word.

Removed 10px and used `$notification-panel-subtitle-height`. Removed comment because `calc` expression is readable.


> Thirdly, did you experiment with vh values for this instead of a calculation using pixels -- as mentioned in my comment in the issue?

Yes. I wasn't able to fix this issue using `vh` values. 
I see height 100% is also used in parent div `common-canvas-items-container` class. Followed the similar height 100% calculation for notification-panel.
Screenshot 2023-10-16 at 2 04 49 PM

tomlyn (approved):

PR #1669, opened 2024-01-18
#1668 Fixed bug setDefaultValues not picking up defaults 2nd call onwards

### LINKED ISSUE #1668: setDefaultValues not picking up defaults 2nd call onwards
Ref: 3639

I see inside of `_addToControlValues()` it can't pick up default because it looks like once we initialize `controlValue`, it isn't `undefined` anymore. So on next call, it is not able to be recognized as needing to be default even when we have it hidden again.

Alson when a default value is set to integer `0` or an empty string `" "`,  these values should be correctly set as default values.

nmgokhale:
Fixes #1668 

Added 2 new properties having default values `0` and `" "`. These will be hidden if the `hide` checkbox is selected.
I don't see any issues with setting `0` and `" "` as default values. 
Screenshot 2024-01-18 at 2 59 39 PM

 
           in the file; or

           it.


matthoward366 (approved):
lgtm

Issue #1713, opened 2024-02-22
Palette categories do not open with keyboard.


srikant-ch5:
While trying to open Palette categories using keyboard content is not loading but accordion caret changes direction.
Ref: #3660

nmgokhale:
Fixed in PR https://github.com/elyra-ai/canvas/pull/1714

PR #1648, opened 2023-12-14
#1646 Panel accordions are not using the carbon standards

srikant-ch5:
Fixes #1646 
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

Issue #1125, opened 2022-07-22
Remove icons from palette when category is closed


tomlyn:
Ref: #3364

Currently, when a category is closed the category's icons remain in the DOM with the display attribute of one of the containing `DIV`s set to `none`.  This can cause problems when inline SVG icons are shown in the palette and on the canvas (after the user has dragged the icon to the canvas to create a new node).  The problem arises when the SVG icon has a `<defs>` section containing elements (such as a `<mask>`) which are referred to using IDs with a a syntax like 
```
style="mask: ur(#xxx)"
``` 
where xxx is the ID of the `<mask>` in the `<defs>` section.

In this instance the icons on the canvas actually refer to the mask element in the inline icon in the palette -- because that is first in the DOM.  When the containing DIV is set to `display: none`   the browsers (Chrome and FF) decise they can no longer access the mask and the icon on the canvas appears incorrectly.

tomlyn:
Fixed: the nodes in a palette category are now removed from the DOM when the category closes and are re-rendered when the category is opened.

PR #1714, opened 2024-02-22
 Display Palette category content using keyboard

srikant-ch5:
Ref: #1713 
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/palette/palette-flyout-content-category.jsx
+++ b/canvas_modules/common-canvas/src/palette/palette-flyout-content-category.jsx
@@ -31,7 +31,8 @@ class PaletteFlyoutContentCategory extends React.Component {
 		this.onMouseOver = this.onMouseOver.bind(this);
 		this.onMouseLeave = this.onMouseLeave.bind(this);
 		this.categoryClicked = this.categoryClicked.bind(this);
-		this.accRef = React.createRef();
+		this.categoryKeyPressed = this.categoryKeyPressed.bind(this);
+		this.setPaletteCategory = this.setPaletteCategory.bind(this);
 	}
 
 	onMouseOver(ev) {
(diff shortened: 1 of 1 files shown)


tomlyn:
Hi @srikant-ch5 Looks like there is a failure in one of the Cypress test.

tomlyn (changes requested):
Hi Srikant 
It looks like there is a double flash of the node icons when opening the palette categories multiple times using the keyboard. This doesn't happen when using the mouse/trackpad.  Can it be fexed?

Large GIF (458x618)

srikant-ch5:
> Hi Srikant It looks like there is a double flash of the node icons when opening the palette categories multiple times using the keyboard. This doesn't happen when using the mouse/trackpad. Can it be fexed?

Hi @tomlyn Thanks for the review. I have updated PR with a fix in which the functionality is working fine as expected. As per Cypress test  AssertionError: Timed out retrying after 4000ms: Expected to find element: `.bx--accordion__item--active`, but never found it. but while testing in browser manually when I open a category using space/enter it is showing bx--accordion__item--active. So I am looking into this.

matthoward366:
With this being a carbon control shouldn't this already be fully accessible?  I wouldn't want to do anything special for this.  btw, we'd have this same problem in common-properties.

srikant-ch5:
> With this being a carbon control shouldn't this already be fully accessible? I wouldn't want to do anything special for this. btw, we'd have this same problem in common-properties.

Hi @matthoward366 Accordion implementation in Palette is different as content is dynamically loaded based on Accordion Title Obj Click and as per description here https://github.com/elyra-ai/canvas/blob/bb63e21cffecb6a73d1ba7d72

35416d22a2c06cc/canvas_modules/common-canvas/src/palette/palette-flyout-content-category.jsx#L184 so when we tab and click space action is triggered on AccordionItem and not on above TitleObj .

I have tested tabbing in Common Properties Accordion https://github.com/elyra-ai/canvas/pull/1648 and it is working fine.

https://github.com/elyra-ai/canvas/assets/16302714/b05acf39-9206-40aa-aa0e-2e999fef71d8

matthoward366:
Those links don't work for me.  Is there a reason they are implemented differently?  Should we change something with the implementation rather then workaround the issue?

matthoward366:
so the issue here is returning `null` instead of content?  If we were to return the content all the time even when close does it work correctly without needing this code?  If that worked, could we return an empty div instead of null?

matthoward366:
@tomlyn It looks like the content is removed from the DOM because of this issue https://github.com/elyra-ai/canvas/issues/1125.  Would this be fixed if we used `uniquifyIDs` for the inline icons?

tomlyn:
I've played with this quite a lot today and can't see why the flash of the icons occurs when rendering the open category. I don't think using uniquifyIDs will help because the same thing happens when displaying non-SVG icons. I wonder if its a bug in the Carbon code and, who knows, might be fixed in Carbon v11. Anyway, this fix is reasonable given the way we are calling the Accordion.

tomlyn (approved):

Issue #1919, opened 2024-05-03
Scroll Alert into view in Common Properties when clicked on an alert from Alerts tab.


srikant-ch5:
Ref: 3771

PR #1920, opened 2024-05-03
#1919 Scroll Alert into view in Common Properties when clicked on an Alert Link

srikant-ch5:
Fixes: #1919 
 
           in the file; or

           it.


srikant-ch5:
Hi @matthoward366 Could you please review this PR and let me know in case of any changes ?

canvas_modules/common-canvas/src/common-properties/components/editor-form/editor-form.jsx
```diff
@@ -89,6 +89,15 @@ class EditorForm extends React.Component {
 		return true;
 	}
 
+	componentDidUpdate() {
+		// Scroll to the selected accordion even when clicked from Alerts tab
+		const activeTabId = this.props.activeTab;
```
matthoward366, line 94: Should this only be set when the activeTab changes?  There probably should be a check that compares the old/new props to make sure the activeTab changed.  If it's different then find the element and scroll to it.

srikant-ch5: > Should this only be set when the activeTab changes? There probably should be a check that compares the old/new props to make sure the activeTab changed. If it's different then find the element and scroll to it.

Hi @matthoward366 since we are enabling multiple tabs to be open at once so there are multiple `activeTab` values. In below scenario if I click on Alert it will scroll to summary-panel but if I click on Alert again it will not scroll in cases if we check props since both are summary-panel.

https://github.com/elyra-ai/canvas/assets/16302714/574517b5-0349-46d4-98af-46c91b5c2846

matthoward366: I understand it works fine but in a `componentDidUpdate` update call usually you want to check if the current props and old props are different before setting something.  I think wrapping this in an if check makes things safer and wouldn't be setting `activeTabId` each time componentDidUpdate is called which is every time there is a prop or state update.

srikant-ch5: Sure Matt. I have updated PR with a condition to check props and check if the alert link is clicked, please let me know if I need to follow any other approach.

canvas_modules/common-canvas/src/common-properties/components/editor-form/editor-form.jsx
```diff
@@ -89,6 +89,18 @@ class EditorForm extends React.Component {
...
 
+	componentDidUpdate(prevProps) {
+		// Scroll to the selected accordion even when clicked from Alerts tab
+		if (this.scrollToAlert || prevProps.activeTab !== this.props.activeTab) {
+			const activeTabId = this.props.activeTab;
+			const activeTabElement = document.getElementsByClassName(`${activeTabId}`);
+			if (activeTabId && activeTabElement.length > 0) {
+				activeTabElement[0].scrollIntoView({ behavior: "smooth" });
+			}
+			this.scrollToAlert = false;
```
matthoward366, line 100: Why do you need `this.scrollToAlert`?  Shouldn't `prevProps.activeTab !== this.props.activeTab` be enough?

srikant-ch5: When I click on alert link then for the first time it is set as activeTab and if I scroll up to Alerts tab again and click then this check fails so I included `this.scrollToAlert`

matthoward366 (approved):
Looks good

PR #1946, opened 2024-05-14
#1945 Fixed multiselect infinite loop after after Carbon v11.57.0 update

### LINKED ISSUE #1945: Fix multiselect infinite loop issue after after Carbon v11.57.0 update
Ref: 3789

Elyra builds are failing since 5/9/2024 after Carbon released new version 11.57.0.

In latest carbon version [v11.57.0](https://github.com/carbon-design-system/carbon/releases/tag/v11.57.0), seeing this error for Multiselect control -
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

Because of this our unit tests are failing due to infinite loop.

Carbon team suggested this solution -
It looks like you could update your `handleOnChange` to compare the `selectedItems` that are in the event with your existing list, and only update the value if they have changed. This would include refactoring to not call  `genSelectOptions` on every render and to used a stored value instead.

nmgokhale:
Fixes #1945 

Initializing `multiSelectDropdown` in render() was causing the infinite loop.
To fix the issue, created a class variable `this.multiSelectDropdown` in constructor and only update its value when either `selectedOptions` have changed or `controlOpts` have changed in case of filterable Multiselect. 

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/controls/multiselect/multiselect.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/controls/multiselect/multiselect.jsx
@@ -38,6 +38,8 @@ class MultiSelectControl extends React.Component {
 		this.state = {
 			shouldRender: true
 		};
+		this.multiSelectDropdown = this.genSelectOptions(props.value);
+		this.selectedOptions = props.value;
 	}
 
 	componentDidMount() {
@@ -45,8 +47,12 @@ class MultiSelectControl extends React.Component {
 	}
 
 	componentDidUpdate(prevProps) {
+		if (!isEqual(this.props.value, prevProps.value)) {
+			this.multiSelectDropdown = this.genSelectOptions(this.props.value);
+		}
 		// only update if filter options have changed. Fixes issue where filter options are updated after value in setProperties
 		if (!isEqual(this.props.controlOpts, prevProps.controlOpts)) {
+			this.multiSelectDropdown = this.genSelectOptions(this.props.value);
 			this.updateValueFromFilterEnum();
 		}
 		this.setShouldRender(prevProps);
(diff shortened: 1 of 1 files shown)


canvas_modules/common-canvas/src/common-properties/controls/multiselect/multiselect.jsx
```diff
@@ -112,11 +118,15 @@ class MultiSelectControl extends React.Component {
 		for (let i = 0; i < evt?.selectedItems?.length; i++) {
 			controlValues.push(evt.selectedItems[i].id);
 		}
+		if (!isEqual(this.selectedOptions, controlValues)) {
```
matthoward366, line 121: Is this check actually needed?  Wouldn't the `handleOnChange` method only be called if values change?

nmgokhale: Carbon team also suggested to add this check -

> It looks like you could update your handleOnChange to compare the selectedItems that are in the event with your existing list, and only update the value if they have changed.

Removing this check causes infinite loop again. 

Infinite loop happens because -

1. We update `this.multiSelectDropdown = this.genSelectOptions(controlValues);` in `handleOnChange()`
2. We call `this.props.controller.updatePropertyValue()` in `handleOnChange()`
3. After the `updatePropertyValue` is complete, `componentDidUpdate()` is called in which we update `this.multiSelectDropdown = this.genSelectOptions(controlValues);` again
4. Because `this.multiSelectDropdown` is updated, again `handleOnChange()` method is called and Steps 1-4 keep happening in loop

matthoward366: I see.  Thanks for the explanation.

Issue #2022, opened 2024-06-17
Arrange Horizontal does not work when we have a detached link.


srikant-ch5:
Ref: 3799

PR #2023, opened 2024-06-17
#2022 Arrange Horizontal does not work when we have a detached link.

srikant-ch5:
Fixes: #2022 
 
           in the file; or

           it.


srikant-ch5:
Hi @tomlyn I can see that the issue #2022  can be solved by changing the spacing to 1.
I have tried to implement dummy node approach mentioned in ticket #3799.
I have done something like below.
```
dagreAutolayout(direction, canvasInfoPipeline) {
		const canvasLayout = this.objectModel.getCanvasLayout();

		let newArr = []; 
		var nodeLinks = canvasInfoPipeline.links.filter((link) => {
			if (link?.trgPos) {
				const dummy_node = this.objectModel.getAPIPipeline().createNode({
					editType: "createNode",
					nodeTemplate: this.dummyNodeTemplate,
					offsetX: link?.trgPos?.x_pos,
					offsetY: link?.trgPos?.y_pos
				});


				newArr.push(dummy_node);

			}

			return link.type === NODE_LINK || link.type === ASSOCIATION_LINK;
		});

		var edges = nodeLinks.map((link) => {
			if ( link?.trgPos) {
				this.objectModel.getAPIPipeline().setLinkProperties(link.id, { trgNodePortId: newArr[0].id })
			}
			return { "v": link.srcNodeId, "w": link.trgNodeId, "value": { "points": [] } };
		});
...
}
```
I think this is not the right approach to create a dummy node do you have any code reference about how to create a dummy node ?

tomlyn:
@srikant 
I added some comments into the issue: https://github.ibm.com/NGP-TWC/wdp-abstract-canvas/issues/3799#issuecomment-82444672

srikant-ch5:
Hi @tomlyn I have included the dummy node ids in nodesData now horizontal layout is working fine.

But I am trying to understand below.

```
After the outputGraph comes back from Dagre, we need to look through the nodes for any with an ID beginning with "temp-src-node" or "temp-trg-node" and then apply its x/y coordinates to the link that the dummy node is for. You will be able to extract the link ID from the dummy node ID by striping off the text prefix. So, for example, if a node coming back from Dagre has ID "temp-src-node-1234" we would look for a link with ID "1234" and overwrite its srcPos with the coordinates provided for that dummy node.
```

the outputGraph returns data format for `nodes` and `edges` in below format which doesn't have `id's` for links.

```
{
    "options": {
        "directed": true,
        "multigraph": false,
        "compound": false
    },
    "nodes": [
        {
            "v": "d7a82866-e652-4949-ba68-494157e217b2",
            "value": {
                "width": 150,
                "height": 75,
                "x": 125,
                "y": 90
            }
        },
        {
            "v": "5fdfd4bd-40ab-43b8-af3f-653645711760",
            "value": {
                "width": 150,
                "height": 75,
                "x": 275,
                "y": 90
            }
        },
        {
            "v": "temp-trg-node-407a23d0-0c02-4c2d-a4d0-3b61da6fa346",
            "value": {
                "width": 150,
                "height": 80,
                "x": 425,
                "y": 90
            }
        }
    ],
    "edges": [
        {
            "v": "d7a82866-e652-4949-ba68-494157e217b2",
            "w": "5fdfd4bd-40ab-43b8-af3f-653645711760",
            "value": {
                "points": [
                    {
                        "x": 200,
                        "y": 90
                    },
                    {
                        "x": 200,
                        "y": 90
                    },
                    {
                        "x": 200,
                        "y": 90
                    }
                ]
            }
        },
        {
            "v": "5fdfd4bd-40ab-43b8-af3f-653645711760",
            "w": "temp-trg-node-407a23d0-0c02-4c2d-a4d0-3b61da6fa346",
            "value": {
                "points": [
                    {
                        "x": 350,
                        "y": 90
                    },
                    {
                        "x": 350,
                        "y": 90
                    },
                    {
                        "x": 350,
                        "y": 90
                    }
                ]
            }
        }
    ],
    "value": {
        "rankDir": "LR",
        "marginx": 50,
        "marginy": 50,
        "nodesep": 80,
        "ranksep": 0,
        "width": 550,
        "height": 180
    }
}
```

tomlyn:
@srikant-ch5 
I don't understand what you are asking.  Since the turnaround in communication is so big it would be better if you asked a more detailed question. For example why are you expecting srcNodes when your example doesn't seem to contain an example of a semi-detached link without a source node. I'll try to reply as best as I can.

From your example above, there is a node in the output graph with a field `v` which contains `temp-trg-node-407a23d0-0c02-4c2d-a4d0-3b61da6fa346`.  So `v` is obviously the node's ID that you set in the nodesData input to dagre. If you did it the way I suggested, it is a target dummy node at the end of a detached link and, again if you did it the way I suggested,   `407a23d0-0c02-4c2d-a4d0-3b61da6fa346` is the link ID of that detached link.  So what you would do is look for a link with ID `407a23d0-0c02-4c2d-a4d0-3b61da6fa346`  and set its `trgPos` field to the x/y coordinates of the node that is in `trgPos` set x to 425 and y to 90.

So you should have gone from this:
```
                                        Link: 40.. 
  Node: d7... -----------> Node: 5f... ---------------> 
```
to add the temporary node:
```
                                         Link: 40.. 
  Node: d7... -----------> Node: 5f... --------------->  Node: temp-trg-node-40...
```
then look for link with ID 40... and update the `trgPos` with x/y coordinates.


By the way, I see in your changes you are updating dummyNodes inside the filter function. That is ugly, just do that processing in a separate loop that loops through the filtered nodes.

srikant-ch5:
HI @tomlyn Thanks for the info. About my above query I was just quoting from your instructions in the ticket (I should have looked into the example data I provided and should have framed it as srcNodes/trgNodes ). Anyway as per our discussion.

> You’d need an additional output array containing the modified links just like movedNodesInfo  contains info about what nodes need to be moved. The additional array would need to be passed into sizeAndPositionObjects()   method in api-pipeline.js.  If you look at that method it does accept a second parameter called linksInfo  which isn’t passed at that moment by the auto layout code.
So you’d need to pass the info in that parameter in the format the method is expecting.  If you look around there will be some other code that calls that method that will give you an idea of what that format is or you can look in the reducer and see how the parameter is being used.

Now its clear to me. Thanks for that.
Regarding my changes I will move out code from filter as it was just something I was testing and separated it afterwards.

srikant-ch5:
Hi @tomlyn

I have created dummy nodes and included in `objectsInfo`.
Here in below reducer it is picking only nodes from `pipeline.nodes` so dummy nodes present in `objectsInfo` are not repositioned. 
https://github.com/elyra-ai/canvas/blob/830bfe34d82541c3dd695cad7b7498cc110d262f/canvas_modules/common-canvas/src/object-model/redux/reducers/nodes.js#L65
To include those nodes as well I have made changes to above reducer to reposition `dummy node` which is not a good approach I think. Could you please suggest any better approach for this ?

tomlyn:
@srikant-ch5 First of all, if you're going to ask me about code that you are working on you need to push that to the PR so I can get a better picture of what you are asking. You really have to help me to help you. 

Second, the dummy nodes should only be created as a convenience in api-pipeline.js.  They are not supposed to be added to the object model because they are not going to appear on the canvas. 

Let's go through the process.  

1. With a canvas **without any** detached links, if the user clicks the auto arrange button the auto-layout code in api-pipeline.js creates input for Dagre that consists of an array of input nodes and an array of input links. Lets call the input nodes 'dagre nodes' and input links 'dagre links'. The dagre nodes and dagre links are different from the actual nodes and links in the object model.  Dagre calculates where the dagre nodes should be, given the type of auto-layout being performed (horizontal or vertical), and returns the set of dagre nodes to common-canvas that have new positions. Common-canvas assigns those positions to the corresponding actual nodes on the canvas - this is done through the reducer.

2. In the case where the flow **does have** one or more detached links we need to trick Dagre into arranging the real nodes in the flow AND any detached links so they appear in acceptable positions. Since Dagre doesn't know anything about detached links, what we do is add a dummy dagre node as if it were at the end of each detached link. This means adding the dummy dagre node to the dagre nodes array and adding a corresponding link to the dagre links array that shows the link going to the dummy dagre node.  Then we ask Dagre to auto-layout the array of dagre nodes and dagre links and Dagre is happy to do that because all it sees is nodes and links and isn't aware that some are dummies and some are 'real'.  

When Dagre returns we need to unravel the return arrays to handle the dummy dagre nodes.  So we loop through the dagre nodes and look for any dagre node with an ID that indicates they are a dummy dagre node. For each dummy dagre node, we update the trgPos (or srcPos) of the corresponding 'real' detached link from the object model with the x/y coordinate of the dummy dagre node returned by Dagre. For all the non-dummy dagre nodes we just update the position of the real nodes from the object model as happens in 1. above. **At this point we have discarded the dummy dagre nodes.**

The new positions for 'real' nodes and 'real' detached links are then applied to the object model using the reducers and the canvas will redraw showing the actual nodes in the positions calculated by Dagre and showing the actual detached links drawn to a position where they **would be** drawn to **if they were attached to real nodes**.

Hopefully you see that the dummy dagre nodes should not be in the object model and therefore there is no need for the reducer to handle them and they are discarded after their position has been used to update the 'real' detached links.

tomlyn:
From  Srikant Chepuri on Slack today:

Hi Craig by referring to your comments I have updated code https://github.com/elyra-ai/canvas/pull/2023/files
i.e if there is a detached link which doesn't have srcNodeId and trgNodeId then add it like below
```
        nodeLinks.forEach((link) => {
            // If there is no srcNodeId create a dummy node id and attach to the detached link
            if (!link.srcNodeId) {
                link.srcNodeId = "temp-src-node-" + link.id;
                dummyNodeIds.push(link.srcNodeId);
            }
            // If there is no trgNodeId create a dummy node id and attach to the detached link
            if (!link.trgNodeId) {
                link.trgNodeId = "temp-trg-node-" + link.id;
                dummyNodeIds.push(link.trgNodeId);
            }
        });
        var edges = nodeLinks.map((link) => {
            return { "v": link.srcNodeId, "w": link.trgNodeId, "value": { "points": [] } };
        });
```
here in edges dummy src/trg will be added to link v/w based on it format.
and then add the nodes to
```
       // Push the dummy Nodes to nodeData so that dagre treats them as nodes attached to detached links
        dummyNodeIds.forEach((dummyNodeId) => {
            nodesData.push({ "v": dummyNodeId, "value": { width: 150, height: 80 } });
        });
```
and created convertGraphToMoveLinks to update links to be passed to reducer.
but detached link end is not moving when layout is changed so I thought its because of reducer code where the nodes positions are actually set.

So just wanted to ask am I missing something here https://github.com/elyra-ai/canvas/pull/2023/files as the detached link end is not moving

tomlyn:
@srikant-ch5 
Your code seems to be 
1. Leaving the temporary node in the `movedNodesInfo` returned from `convertGraphToMovedNodes`:

    image


3. Leaving the trgNodeId in the semi-detached link in `movedLinksInfo` returned from `convertGraphToMoveLinks`. A detached link should only have a position field (containing the x/y coordinate of the detached end of the link) and no Node ID field for that end of the link. 

    image

srikant-ch5:
@tomlyn as per above comments, I have updated code 
1. To remove temp dummy node from `movedNodesInfo`.
Screenshot 2024-06-25 at 8 21 59 PM

2. To remove `trgNodeId` or `srcNodeId` for detached links in `movedLinksInfo `.
Screenshot 2024-06-25 at 8 23 58 PM

tomlyn:
@srikant-ch5 I'm not sure what you message is trying to say. The screen shot seems to show it not working correctly.  Does that mean you are still working on the issue? It really would be helpful if in your messages you could be clearer about what you are asking/telling me.

srikant-ch5:
Hi @tomlyn Yes its not working as expected.

srikant-ch5:
@tomlyn
As per above steps I have included dummy trg node at semi-detached link trgNode.
Below is the inputgraph.nodes which we are sending to dagre (which includes dummy node).
```
[
    {
        "v": "93def4b1-c82f-4ca4-86c2-a7a2a394ddaa",
        "value": {
            "width": 150,
            "height": 75,
            "x": 125,
            "y": 90
        }
    },
    {
        "v": "4076d831-7d28-4fe0-8413-9b5837e5c902",
        "value": {
            "width": 150,
            "height": 75,
            "x": 275,
            "y": 90
        }
    },
    {
        "v": "temp-trg-node-3e69e340-1146-45c3-af98-8e61a9deb57f",
        "value": {
            "width": 150,
            "height": 80,
            "x": 425,
            "y": 90
        }
    }
]  
```

below is the outputgraph.nodes returned from dagre.
```
[
    {
        "v": "93def4b1-c82f-4ca4-86c2-a7a2a394ddaa",
        "value": {
            "width": 150,
            "height": 75,
            "x": 125,
            "y": 90
        }
    },
    {
        "v": "4076d831-7d28-4fe0-8413-9b5837e5c902",
        "value": {
            "width": 150,
            "height": 75,
            "x": 275,
            "y": 90
        }
    },
    {
        "v": "temp-trg-node-3e69e340-1146-45c3-af98-8e61a9deb57f",
        "value": {
            "width": 150,
            "height": 80,
            "x": 425,
            "y": 90
        }
    }
]
```

My question is here as we can see x/y cordinates doesn't change in outputgraph it remains same as in inputgraph. In `convertGraphToMoveLinks` we are passing same nodes so the semi-detached link end stays even tough if we attach x_pos and y_pos of above dummy node its remains at same point is what I believe. Any suggestion from your side to solve this ?

tomlyn:
@srikant-ch5 Thanks for the extra information. 


https://github.com/elyra-ai/canvas/assets/15114439/70a31f9d-9cb6-4e51-a8a6-4c9485dea44e


https://github.com/elyra-ai/canvas/assets/15114439/c3304223-afea-4f75-b895-3e9772d3e49a

srikant-ch5:
Thanks @tomlyn for pointing out the mistake.
All this while I was debugging nodes and its reducer thinking node is not moving (because visually on canvas when we drag a node it also leads to movement of the attached link and not vice versa so I was looking in wrong direction all together). 
Now its fixed and detached link is moving as per layout.

https://github.com/elyra-ai/canvas/assets/16302714/c62b35f8-29c5-4584-b10f-673c682367f0

srikant-ch5:
Hi @tomlyn I have updated PR to handle a fully detached link auto layout.
Please refer below recording and let me know if the layout is fine.

https://github.com/elyra-ai/canvas/assets/16302714/359576ab-0c5d-48f2-ba62-77cbdfbf1fc7

Also I have changed below variable from `array` to an `object` because in the code below its being used as an objectm hope this is fine.
https://github.com/elyra-ai/canvas/blob/830bfe34d82541c3dd695cad7b7498cc110d262f/canvas_modules/common-canvas/src/object-model/api-pipeline.js#L917

canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -936,6 +958,73 @@ export default class APIPipeline {
...
+		const movedLinksInfo = {};
+		const linksPos = [];
+
+		const nodes = outputGraph.nodes;
+
+		nodes.forEach((node) => {
+			// Record the x and y position of dummy node to be assigned to the detached link.
+			// A fully detached link can have both temp-src-node and temp-trg-node
+			if (node.v.startsWith("temp-src-node-")) {
+				const dummyNodeId = node.v.split("temp-src-node-")[1];
```
tomlyn, line 975: I think this is the linkId -- isn't it?

canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -936,6 +958,73 @@ export default class APIPipeline {
...
+
+				linksPos[dummyNodeId] = {
+					...linksPos[node.v.split("temp-src-node-")[1]],
+					srcPos: {
+						x: node.value.x,
+						y: node.value.y
+					}
+				};
```
tomlyn, line 983: This seems a little over-complicated. It seems like you are getting the linkId again 
```
node.v.split("temp-src-node-")[1]
```
which you already got in the line above -- and then referencing into an element of the link array that will not have anything in it. (Although in the next section down for target Pos it may have something in it if we are handling a fully-detached link).

canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -936,6 +958,73 @@ export default class APIPipeline {
...
+				linksPos[dummyNodeId] = {
+					...linksPos[node.v.split("temp-trg-node-")[1]],
+					trgPos: {
+						x: node.value.x,
+						y: node.value.y
+					}
+				};
+			}
+		});
+
```
tomlyn, line 997: I don't think the code below is totally necessary. If you look in the `links.js` reducer at the code for case "SIZE_AND_POSITION_OBJECTS"  you'll see the only fields it needs for each element of `movedLinksInfo` is {id, srcPos, trgPos }.  So the  `movedLinksInfo` doesn't need to be an array of links with all their fields but just an array of elements with the above 3 fields which is basically what the linksPos array is.

tomlyn (changes requested):

tomlyn:
@srikant-ch5
 I think I answered your questions in my review comments. Let me know if I missed anything.

Don't also forget we will need some tests for this changed function. I'd suggested adding a test case to the Cypress `autoLayoutDagre.cy.js` file.  You can use the existing `detachedLinksCanvas.json` pipeline flow file for the test. You will have to:
*  switch  `selectedLinkSelection` to "Detachable" in the test to get the detached links to appear, and
* specify the SingleLeftBarArray toolbar like the other tests and
* simulate a click on the horizontal auto layout buttons and 
* do an undo and 
* do a redos 
* do a click on the vertical auto layout buttons and 
* do an undo and 
* do a redos 
* and verify that the nodes and links are in the correct places after each step.

srikant-ch5:
Hi @tomlyn thanks for the feedback.
I have updated PR as per your comments and have also included cypress test, please review.
Only thing I haven't incorporated yet is below 

> I was wondering why, in your video of the fully detached link, the ends of the link are not exactly underneath the nodes after auto-layout. I think that is because we are using these hard-coded values for the width and height which .....

Is it okay if we track this under a new issue and discuss more on this under that ?

tomlyn:
> Is it okay if we track this under a new issue and discuss more on this under that ?

No it's not. This is part of the fix otherwise it won't work correctly.  I don't understand why you are even asking. What is the reason for doing it separately? I've literally typed in the code you need to use so you just need to put it in your code and test it out.

srikant-ch5:
> > Is it okay if we track this under a new issue and discuss more on this under that ?
> 
> No it's not. This is part of the fix otherwise it won't work correctly. I don't understand why you are even asking. What is the reason for doing it separately? We need to get the whole thing working and then merge the code. There's no advantage to have a separate issue as far as I can see.

Ok Craig Sorry for that. I was just asking because I will be OOO for next two weeks. I included the code you suggested but since it was not working I thought to continue on it once I get back. 
Could you please review that code and let me know ?

tomlyn:
That's OK -- you just need to let me know why you are asking.  Since you are out from now I can take over this item.  If I can't figure it out I'll just merge your code as is. Have a great vacation (holiday)!

PR #2120, opened 2024-08-26
NGP-TWC/wml-canvas-planning#3851 Added icon support for dropdown menus

PRINCESANCHEZ:
Fixes: #2192 

           in the file; or

           it.


canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,8 +266,26 @@ class DropDown extends React.Component {
 					disabled={this.props.state === STATES.DISABLED || this.disableEmptyListDropdown}
 					type="default"
 					items={dropDown.options}
+					itemToElement={(item) => (item ? (
+						<span className="icon-support" style={{ color: "black" }}>
+							<span>{ item.value }</span>
+							<Password style={{ float: "right" }} />
```
caritaou, line 272: This is a good POC. But we will need this to be passed in by the consumer application, specified in the uihints. The default will be no icon. 

A new hint will need to be added here: https://github.com/elyra-ai/pipeline-schemas/blob/main/common-pipeline/operators/uihints-v3-schema.json#L209

This is what defines a `oneofselect` dropdown control:
image

The new hint will need to be able to specify an image for whatever dropdown option that needs an image shown. Or maybe a callback to the application to give us the image(s).

From the use case, seems like only a few options will require a single image. @matthoward366 Thoughts on how this should be implemented?

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,8 +266,26 @@ class DropDown extends React.Component {
...
+						</span>
+					) : (
+						" "
+					))
+					}
 					onChange={this.handleChange}
 					selectedItem={dropDown.selectedOption}
+					itemToString={(item) => (item ? (
+						<span className="icon-support" style={{ color: "black", display: "flex", justifyContent: "space-between" }}>
```
caritaou, line 281: We dont specify inline styles unless its absolutely necessary. We set styles in scss files using class names that are unique to this class so that there are no chances of any classname overlaps with those of the application

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
matthoward366: This was discussed and decided to add a new callback.  We could add something like a `iconCallback` or something similar.  When the call is called we'd need to pass the propertyId, data, and specific enum.  The enum should be optional in case we want to reuse the callback in other places.

caritaou: @PRINCESANCHEZ Take a look at this documentation on how callbacks will be passed in https://elyra-ai.github.io/canvas/04.02-callbacks. Once implemented, these docs will need to be updated to include the new callback as well

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -303,7 +317,8 @@ DropDown.propTypes = {
 		PropTypes.object
 	]), // pass in by redux
 	messageInfo: PropTypes.object, // pass in by redux
-	readOnly: PropTypes.bool
+	readOnly: PropTypes.bool,
+	iconCallBack: PropTypes.func
```
matthoward366, line 321: How dos this work?  There is nothing passing this callback into this component. 

A couple things are needed
1. Add the callback into https://github.com/elyra-ai/canvas/blob/main/canvas_modules/common-canvas/src/common-properties/common-properties.jsx#L257.  This is an object of callback for common-properties.
2. Once the callback is added you can use the properties controller to access the callback.
3. There needs to be a paramDef updated and the test harness updated to test this callback.  Possibly a toggle that turns it on/off.
4. Unit tests.

canvas_modules/common-canvas/src/common-properties/common-properties.jsx
```diff
@@ -266,7 +266,8 @@ CommonProperties.propTypes = {
 		validationHandler: PropTypes.func,
 		titleChangeHandler: PropTypes.func,
 		propertiesActionLabelHandler: PropTypes.func,
-		tooltipLinkHandler: PropTypes.func
+		tooltipLinkHandler: PropTypes.func,
+		iconCallBack: PropTypes.func
```
matthoward366, line 270: Maybe rename this to propertyIconHandler.  We already have a buttonIconHandler so this kind of matches that.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
matthoward366: There is still something missing here with `const { iconCallBack } = this.props;`.  This doesn't get passed in today.

canvas_modules/harness/test_resources/parameterDefs/oneofselect_paramDef.json
```diff
@@ -45,6 +45,7 @@
         "blue",
         "purple"
       ],
+      "enableIconCallBack": true,
```
matthoward366, line 48: I don't think we'd add anything to the paramDef.  This would be a change in the test harness to implement the callback above.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,6 +280,20 @@ class DropDown extends React.Component {
 					disabled={this.props.state === STATES.DISABLED || this.disableEmptyListDropdown}
 					type="default"
 					items={dropDown.options}
+					itemToString={(item) => (item ? (
```
matthoward366, line 283: Why is `itemToString` and `itemToElement` both needed and both the same code?  I would think `itemToString` would be a string?

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -87,6 +88,20 @@ class DropDown extends React.Component {
 		return selectedOption;
 	}
 
+	testIconCallBack(propertyId, enumValue) {
```
matthoward366, line 91: This should probably be getItemIcon or something like that.

canvas_modules/common-canvas/src/common-properties/common-properties.jsx
```diff
@@ -266,7 +268,9 @@ CommonProperties.propTypes = {
 		validationHandler: PropTypes.func,
 		titleChangeHandler: PropTypes.func,
 		propertiesActionLabelHandler: PropTypes.func,
-		tooltipLinkHandler: PropTypes.func
+		tooltipLinkHandler: PropTypes.func,
+		propertyIconHandler: PropTypes.func,
+		iconCallBack: PropTypes.func
```
matthoward366, line 273: Why do we have 2 new callbacks?

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,6 +280,20 @@ class DropDown extends React.Component {
...
 					items={dropDown.options}
+					itemToString={(item) => (item ? (
+						<div className="item-name">
+							<div>{ item.label }</div>
+							<div className="custom-icon-label">{this.testIconCallBack()}</div>
+						</div>
+					) : ""
+					)}
+					itemToElement={(item) => (item ? (
+						<div data-id={`properties-${this.props.propertyId}`} className="item-name">
```
matthoward366, line 291: data-id isn't usually set on each specific value.  That seems wrong.  `item-name` also seems too generic.  We should at least add properties-...

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
matthoward366: I don't think this is actually used.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.scss
```diff
@@ -23,6 +23,20 @@
 	}
 
 }
+
+.cds--dropdown{
```
matthoward366, line 27: This shouldn't use a carbon class at the top level.  This would potentially style components outside of our app.

canvas_modules/harness/src/client/App.js
```diff
@@ -27,6 +27,7 @@ import { FormattedMessage, IntlProvider } from "react-intl";
 import { forIn, get, has, isEmpty, isEqual } from "lodash";
 import classNames from "classnames";
 import { v4 as uuid4 } from "uuid";
+import { Password } from "@carbon/icons-react";
```
matthoward366, line 30: Why are we using a control instead of an icons?

canvas_modules/common-canvas/src/common-properties/common-properties.jsx
```diff
@@ -252,7 +253,8 @@ CommonProperties.propTypes = {
 		convertValueDataTypes: PropTypes.bool,
 		showRequiredIndicator: PropTypes.bool,
 		showAlertsTab: PropTypes.bool,
-		locale: PropTypes.string
+		locale: PropTypes.string,
+		iconSwitch: PropTypes.bool
```
matthoward366, line 257: Why do we need this?  If they are passing in the `propertyIconHandler` they shouldn't also need to set this config option.

canvas_modules/harness/test_resources/parameterDefs/oneofselect_paramDef.json
matthoward366, line 48: Looks like this is still in the PR?

canvas_modules/common-canvas/src/common-properties/common-properties.jsx
PRINCESANCHEZ: Forgot to remove this one from previous commit

canvas_modules/harness/src/client/App.js
matthoward366: Ignore.  This is fine.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.scss
```diff
@@ -55,3 +57,12 @@
 		}
 	}
 }
+
+.properties-dropdown-label {
+	display: flex;
+	justify-content: space-between;
+	.custom-icon {
+		padding-left: 160px;
```
matthoward366, line 65: Where did you get this number?  There should be a way to place the icon at the end of that is what is desired.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,6 +280,13 @@ class DropDown extends React.Component {
 					disabled={this.props.state === STATES.DISABLED || this.disableEmptyListDropdown}
 					type="default"
 					items={dropDown.options}
+					itemToString={(item) => (item ? (
+						<div className="properties-dropdown-label">
+							<div>{ item.label }</div>
+							<div className="custom-icon">{this.getItemIcon()}</div>
```
matthoward366, line 286: What if there is no icon returned?  I don't think we should add the div if the callback returns a null or undefined.

docs/pages/04.02-callbacks.md
```diff
@@ -114,6 +114,23 @@ buttonIconHandler(data, callbackIcon) {
...
+-data: an object that consists of
+  - propertyId: of the dropdown that was selected
+  - enumValue: of the dropdown that was selected
+
+```js
+propertyIconHandler(data, callbackIcon) {
+  const propertyId = { "name": "oneofselect" };
+	const { iconSwitch } = this.state;
+	if (iconSwitch === true && data.propertyId.name === propertyId.name) {
+		callbackIcon(<Password />);
```
matthoward366, line 129: I would rename `Password` to `Icon` or something that makes it more clear it's an icon users should pass in.

docs/pages/04.02-callbacks.md
```diff
@@ -114,6 +114,23 @@ buttonIconHandler(data, callbackIcon) {
 }
 ```
 
+### propertyIconHandler()
+Called when a user wants to pass in a specific icon to a dropdown menu. The propertyIconHandler expects a Carbon Icon jsx object as the return value from the callback. This is used to display the Carbon icon in the dropdown menu. The propertyIconHandler provides the following data:
```
matthoward366, line 118: `Carbon Icon jsx object`

Is this true?  Can't they pass in any object?

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -265,6 +280,13 @@ class DropDown extends React.Component {
...
+						</div>
+					) : ""
```
matthoward366, line 288: Does this fallback to the label in this case?  It shouldn't show the raw value.

docs/pages/04.02-callbacks.md
```diff
@@ -114,6 +114,23 @@ buttonIconHandler(data, callbackIcon) {
 }
 ```
 
+### propertyIconHandler()
+Called when a user wants to pass in a specific object to a dropdown menu. The propertyIconHandler expects a Carbon jsx object as the return value from the callback. This is used to display the Carbon icon in the dropdown menu. The propertyIconHandler provides the following data:
```
matthoward366, line 118: ` Carbon jsx object`
Should just be `jsx object`

`This is used to display the Carbon icon in the dropdown menu`
Should be changed to 
`This is used to display the jxx object in the dropdown menu`

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
PRINCESANCHEZ: I ended up using item.label, since item.value failed multiple tests

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -87,6 +88,20 @@ class DropDown extends React.Component {
 		return selectedOption;
 	}
 
+	getItemIcon(propertyId, enumValue) {
```
matthoward366, line 91: `propertyId` is never used and `enumValue` is actually passed into the function below.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -212,6 +227,13 @@ class DropDown extends React.Component {
 			"clear.selection": formatMessage(this.reactIntl, MESSAGE_KEYS.DROPDOWN_TOOLTIP_CLEARSELECTION)
 		};
 
+		const renderItem = (item) => (item ? (
```
matthoward366, line 230: It's generally better to have this be it's own function.  Having it inline recreate the function x times.

matthoward366 (changes requested):

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -87,6 +89,20 @@ class DropDown extends React.Component {
 		return selectedOption;
 	}
 
+	getItemIcon(enumValue) {
```
matthoward366, line 92: enumValue isn't passed into the function so this wouldn't be passed down.

Can you also add a test case to make sure this actually works correctly?  You should validate the propertId and enumValue are correct.  Might also be worth having 2 different icons in the example to show this works.

matthoward366 (approved):
Looks good

matthoward366:
@PRINCESANCHEZ This need an issue from https://github.com/elyra-ai/canvas/issues linked. The current link isn't valid.

Issue #2137, opened 2024-09-03
Enable common properties right flyout resize using drag


srikant-ch5:
Ref: 3797

PR #2138, opened 2024-09-03
#2137 Enable right flyout resize using drag

srikant-ch5:
Fixes: #2137
 
           in the file; or

           it.


srikant-ch5:
Hi @matthoward366 I have removed test cases related to resize buttons as they are replaced with drag, is this okay ?

https://github.com/user-attachments/assets/d2142018-00c9-45bd-8823-239e338792c6

matthoward366:
The design expected to keep the buttons as well as having the dragging option.  That way customer can quickly increase the size.

This also needs to be behind a feature flag since not all customers would want this.

tomlyn:
@matthoward366 
> The design expected to keep the buttons as well as having the dragging option.

I think Ana suggests this behavior would replace the buttons. Which is why, I assume @srikant-ch5 has done it this way. I've queried that with Ana

@srikant-ch5 I think we ought to implement this is a different way. Right now the user can dynamically size the top panel and bottom panel by drag/drop and that code is in Common Canvas. I think it makes sense therefore that we should have similar code for the right-flyout (and also probably the left flyout too) inside Common Canvas - not in Common Properties. That way applications can get the sizing behavior even if they are not using Common-Properties. As Matt points out, not all applications will want to the drag/drop sizing behavior for the right-flyout so we should have a feature flag for that (an one for the left flyout too if we implement drag/drop sizing for that too).

srikant-ch5:
Hi @tomlyn Yes correct, I checked with her regarding few other queries.

Sure I will revert the changes in Common Properties and will try to implement this in Common Canvas.

srikant-ch5:
@tomlyn I Included the sizing behaviour in Common Properties because I tried implementing it CommonCanvas Right flyout and when I expand the panel content dosen't grow/shrink with it, will try to look for a solution for this.


https://github.com/user-attachments/assets/a06cfa06-fac8-4e9b-b142-3d6f95adcf16

srikant-ch5:
Hi @matthoward366 is it okay if we set the width of properties right flyout like below in `properties-main.jsx` since we are implementing this in common canvas and the  properties width should also be applied on cc right lfyout.

https://github.com/elyra-ai/canvas/blob/0baa3c0b77db566f80b8718e81c3df3fccf0f0cf/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx#L429

and in cc-right-flyout like below 

https://github.com/elyra-ai/canvas/blob/ddc8c53e901e7515fe1fcd7b889fd2e4b4870d4e/canvas_modules/common-canvas/src/common-canvas/cc-right-flyout.jsx#L58

I haven't changed anything related to panel sizes in `properties-main.scss`

Currently it behaves like blow to enable extend upto 70% of available width and minimum width. Please let me know your thoughts on this or if you want to suggest any other approach to this. Above changes are in PR.

https://github.com/user-attachments/assets/31ec0429-1add-499f-a44c-f1199a9135d1

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -87,6 +87,12 @@ class PropertiesMain extends React.Component {
 			showPropertiesButtons: true,
 			editorSize: editorSize
 		};
+		this.flyoutWidth = {
```
matthoward366, line 90: I'm not seeing this being used.  Am I missing something?

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -419,11 +426,19 @@ class PropertiesMain extends React.Component {
 		this.setState({ showPropertiesButtons: state });
 	}
 
+	updateRightFlyoutWidth(size) {
```
matthoward366, line 429: This seems like the wrong place for setting this.  Why is this needed here?

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.scss
```diff
@@ -58,6 +59,7 @@ $properties-resize-button-size: $spacing-06;
 	padding: $spacing-01;
 	justify-content: center;
 	align-items: center;
+	z-index: 11;
```
matthoward366, line 62: Why is this needed?  This might cause other issues.

canvas_modules/common-canvas/src/common-canvas/cc-right-flyout.jsx
```diff
@@ -19,10 +19,59 @@ import { connect } from "react-redux";
...
+	}
+
+	onMouseMoveX(e) {
+		if (e.clientX) {
+			const newWidth = this.startWidth + (this.posX - e.clientX);
+			this.commonCanvasRightFlyout.style.width = `${this.limitWidth(newWidth)}px`;
+
+			// In case of Common Properties remove hardcoded width to allow content to take full width.
+			const righFlyoutChildNode = this.commonCanvasRightFlyout.childNodes[1];
+			if (righFlyoutChildNode && righFlyoutChildNode.className.includes("properties-right-flyout")) {
```
matthoward366, line 57: This seems wrong since it's mixing common properties and common-canvas.

canvas_modules/common-canvas/src/common-canvas/cc-right-flyout.jsx
```diff
@@ -35,7 +84,12 @@ class CommonCanvasRightFlyout extends React.Component {
 				? "right-flyout-panel under-toolbar"
 				: "right-flyout-panel";
 			rightFlyout = (
-				<div className={rfClass}>
+				<div className={rfClass} id="right-flyout-panel" ref={ (ref) => (this.commonCanvasRightFlyout = ref) }>
```
matthoward366, line 87: This shouldn't have an id of `right-flyout-panel`.  We should try not to use ids in common-canvas or common-properties.  If we do, then they need to be unique for that canvas since we can have multiple instances in the DOM.

srikant-ch5: Ok, I have removed this now.

canvas_modules/common-canvas/src/common-canvas/cc-right-flyout.jsx
srikant-ch5: Thanks, removed this code from CC.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
srikant-ch5: Since we are trying to implement resize in `cc-right-flyout` which will resize `right-flyout-panel` as per drag width. So when we click on panel resize it will only resize the common properties content and will not expand/shrink the right-flyout-panel so to enforce width on CC right-flyout I have included this. Please let me know if you can suggest another way of doing this.

Without enforcing width to right-flyout in properties it will look something like below, where the properties content width is set as per custom classnames but flyout width is not reset.

Screenshot 2024-09-23 at 12 19 42 PM

Also could you please where can I find any reference to feature flag in the code ?

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
srikant-ch5: This is being used in updateRightFlyoutWidth to enforce width.

tomlyn:
@srikant-ch5 Is this ready for review?

srikant-ch5:
> @srikant-ch5 Is this ready for review?

Hi @tomlyn I have added a comment above https://github.com/elyra-ai/canvas/pull/2138#discussion_r1770846502 and without that code the right-flyout-panel is unable to reset its width when custom width is set in common properties ( custom classnames in  are still set as per size small, medium, etc ..). I have changed this PR to Ready for Review.

matthoward366:
This isn't correct.  The button should be over the line not the other way around.
Screenshot 2024-10-08 at 3 30 47 PM

matthoward366:
A few things I noticed
1. The left side of the expand button is missing a border on hover
2. When shrinking the panel the accordions sometime get cutoff and sometime resize oddly.
3. The documentation page is showing blank http://localhost:3001/#/properties


https://github.com/user-attachments/assets/c19d77e3-e279-4a4b-aacd-702da0ba8886

srikant-ch5:
> A few things I noticed
> 
> 1. The left side of the expand button is missing a border on hover
> 2. When shrinking the panel the accordions sometime get cutoff and sometime resize oddly.
> 3. The documentation page is showing blank http://localhost:3001/#/properties

Thanks for the feedback. Resolved these issues in the latest commit.

srikant-ch5:
Closing this PR due to DCO issue opened a new PR instead. https://github.com/elyra-ai/canvas/pull/2224

Issue #2141, opened 2024-09-05
Latest version of Chrome debugger causing redux to run slowly in test harness


tomlyn:
Ref: #3869

PR #2142, opened 2024-09-05
#2141 Latest version of Chrome debugger causing redux to run slowly i…

tomlyn:
…n test harness

Fixes: #2141 

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/object-model/redux/canvas-store.js
+++ b/canvas_modules/common-canvas/src/object-model/redux/canvas-store.js
@@ -80,12 +80,14 @@ export default class CanavasStore {
 			toppanel: { }
 		};
 
-		if (typeof window !== "undefined") {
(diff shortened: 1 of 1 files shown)


caritaou (approved):

matthoward366 (commented):
Should this be changed in common-properties as well?

tomlyn:
@matthoward366 
I don't know if is affects common-properties or not. It just started happening for me when I updated to the latest Chrome and ran Common Canvas with the debugger open. I get the message below in the console (which I think were were getting before) and thereafter all interactions run slowly.  The peculiar thing is it happens when I open the test harness right-flyout for the first time. So that is before I have done anything substantial in Common Canvas such as open a sample app or select a test flow or palette to display, etc. So at that point there shouldn't be any objects that are very big in Redux. 

image

I was hoping it would go away with a future update of Chrome which is why I left the code there but commented out. To be honest though, even before the problem, I rarely used the Redux debug tools so it's no great loss.

matthoward366:
ok, we can leave it for now unless we see issues.

Issue #2149, opened 2024-09-13
Action Button width Issue for Custom Editor Size


srikant-ch5:
Ref: 3860

PR #2150, opened 2024-09-13
#2149 Action Button width Issue for Custom Editor Size

srikant-ch5:
Fixes: #2149
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
@@ -570,13 +570,16 @@ class PropertiesMain extends React.Component {
 					{editorForm}
 				</PropertiesModal>);
 			}
+
+			// If Custom width is set
+			const customWidth = this.propertiesController.getForm().pixelWidth;
 			const className = classNames("properties-wrapper",
 				{
 					"properties-right-flyout": this.props.propertiesConfig.rightFlyout,
 					"properties-light-enabled": this.props.light,
 					"properties-light-disabled": !this.props.light
 				},
-				`properties-${this.state.editorSize}`);
+				customWidth ? "properties-custom-width" : `properties-${this.state.editorSize}`);
 			const overrideSize = this._getOverrideSize();
 			let overrideStyle = null;
 			if (overrideSize !== null) {


canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -570,13 +570,22 @@ class PropertiesMain extends React.Component {
 					{editorForm}
 				</PropertiesModal>);
 			}
+
+			// If Custom width is set adjust classname to format action buttons
+			const customWidth = this.propertiesController.getForm().pixelWidth;
```
matthoward366, line 575: @srikant-ch5 is there an example where I can see this working?  Can't you just set the class based on `overrideSize` being set?  

```
			if (overrideSize !== null) {
				overrideStyle = { width: overrideSize + "px" };
			}
```

srikant-ch5: Hi @matthoward366 If I include it under `overrideSize` condition below test conditions are failing.
- When enableResize=true and editor_size=small and pixel_width min and max are set resize button should be rendered
- When enableResize=true and editor_size=small and pixel_width min and max are the same the resize button should not be rendered
- When enableResize=true and editor_size is omitted and pixel_width min and max are the same the resize button should not be rendered
https://github.com/elyra-ai/canvas/actions/runs/10934918203/job/30355802106?pr=2150

srikant-ch5:
@matthoward366 For testing I am setting `pixel_width` and `editor_size` in `action_paramDef.json` and now it behaves like below 


https://github.com/user-attachments/assets/7787e9de-7d9a-4bbd-82e4-60357cf921c1

Previously it was behaving in the following way.


https://github.com/user-attachments/assets/76705383-1844-4d07-830d-f6d414b70765

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
matthoward366: Any reason not to update the tests?  In all the above cases the width is going to by the `overrideSize` and not use the className to set the size.

srikant-ch5: I have updated tests and included classname in overrideSize condition.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -570,18 +570,25 @@ class PropertiesMain extends React.Component {
 					{editorForm}
 				</PropertiesModal>);
 			}
+
+			let propertiesSizeClassname = `properties-${this.state.editorSize}`;
+
+			const overrideSize = this._getOverrideSize();
+			let overrideStyle = null;
+			if (overrideSize !== null) {
+				if (this.state.editorSize === Size.SMALL) {
```
matthoward366, line 579: Why do we need this check?  Can't we juts always set this for overrideSize?  if there is a specific reason for this check then we should have a comment here.

srikant-ch5: Since the issue was arising due to `properties-small` classname in below code.

https://github.com/elyra-ai/canvas/blob/b8c2388c921398c4e4aecc9596dd45fa078a570b/canvas_modules/common-canvas/src/common-properties/components/title-editor/title-editor.scss#L42

I have set custom classname only when editor size is small.


https://github.com/user-attachments/assets/a2100ffa-3e3a-4189-8be9-d71a31c5f4e1

If I remove that condition classname won't switch.


https://github.com/user-attachments/assets/b3f4eb53-0593-4a42-9304-39705b0fe276

srikant-ch5: Sure. I have updated the code and also tests to include new classname `properties-custom-size`.

Issue #2199, opened 2024-10-09
Allow expression text editor to be expandable


srikant-ch5:
Ref: 3883

PR #2200, opened 2024-10-09
#2199 Allow expression text editor to be expandable

srikant-ch5:
Fixes: #2199
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
+++ b/canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
@@ -48,7 +48,7 @@ $operator-margin: 50%;
 
 	.elyra-CodeMirror {
 		resize: vertical;
-    	overflow: auto !important;
+		overflow: auto !important;
 		.cm-editor {
 			min-height: inherit;
 			resize: vertical;
--- a/canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
+++ b/canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
@@ -47,12 +47,13 @@ $operator-margin: 50%;
 	margin-bottom: 0;
 
 	.elyra-CodeMirror {
+		/* Allow codemirror container to resize on drag */
 		resize: vertical;
-		overflow: auto !important;
+		overflow: auto;
 		.cm-editor {
-			min-height: inherit;
-			resize: vertical;
-			overflow: auto !important;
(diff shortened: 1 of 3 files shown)


canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
```diff
@@ -47,7 +47,12 @@ $operator-margin: 50%;
 	margin-bottom: 0;
 
 	.elyra-CodeMirror {
+		resize: vertical;
```
matthoward366, line 50: Why do we need to set these stylings in both classes?

matthoward366: Also, why do we need to set `!important`.  If there is a good reason please add a comment.  We should also add a comment as to why this styling was added.

srikant-ch5: Sure @matthoward366 I have updated styling with comments. Please let me know if I need to change these.

I have included resize styling in elyra-CodeMirror which is the outer container so that the container grows along the text editor without this styling it is behaving like below.


https://github.com/user-attachments/assets/5cf65ef6-3001-4397-a196-79bda81be832

Similarly if I don't apply resize styling on inner `cm-editor` it would behave something like below where the editor doesn't resize.

https://github.com/user-attachments/assets/ed682fdd-5904-4115-93bd-7a9906f4226c

canvas_modules/common-canvas/src/common-properties/controls/expression/expression-builder/expression-builder.scss
```diff
@@ -47,7 +47,13 @@ $operator-margin: 50%;
 	margin-bottom: 0;
 
 	.elyra-CodeMirror {
+		/* Allow codemirror container to resize on drag */
+		resize: vertical;
+		overflow: auto;
 		.cm-editor {
+			min-height: inherit; /* Inherit min-height to restrict resize below certain width  */
+			resize: vertical; /* Allow codemirror texteditor to resize on drag */
```
matthoward366, line 55: I removed the resize and overflow here and it works correctly.
https://github.com/user-attachments/assets/5864a04a-7ede-4920-b704-1bb9efe474c1

srikant-ch5: Yes, its working without that I have updated the PR. Thanks.
For some reason it was behaving like above recording now when I restart the browser its working fine.

matthoward366 (approved):

PR #2224, opened 2024-10-24
#2137 Resize Right Flyout using drag

srikant-ch5:
Fixes: #2137
 
           in the file; or

           it.


srikant-ch5:
Old PR: https://github.com/elyra-ai/canvas/pull/2138

srikant-ch5:
Hi @matthoward366 Is it ok to include `properties-right-flyout-container` the way I included it in this PR or will it break any functionality in the consuming applications ? 
I included it to align the resize button so that content spreads out in the properties-panel.
( I will include comments and will refactor the code in following commits)

matthoward366:
@nmgokhale can you have a look?

srikant-ch5:
@tomlyn Could you please let me know whether `cc-rightflyout` code is ok ? I have tried to keep it just like `cc-bottompanel` with few changes like using `ref` to limit panel to 70%.

canvas_modules/common-canvas/src/common-canvas/cc-panels.jsx
```diff
@@ -138,7 +138,7 @@ class CommonCanvasPanels extends React.Component {
 				containingDivId={this.props.containingDivId}
 			/>
 		);
-		const rightFlyout = (<CommonCanvasRightFlyout />);
+		const rightFlyout = (<CommonCanvasRightFlyout containingDivId={this.props.containingDivId} canvasController={this.props.canvasController} />);
```
tomlyn, line 141: Can't this code be structured the same way as we do it for left flyout below? That is, package it in a function; return null if the panel is not open etc.  I'd like the code that handles the left flyout and the code that handles the right flyout be as much as possible structured in the same way.

canvas_modules/common-canvas/src/common-canvas/cc-right-flyout.jsx
```diff
@@ -19,24 +19,73 @@ import { connect } from "react-redux";
...
+			const maxWidth = MAX_WIDTH_EXTEND_PERCENT * totalAvialableWidth;
+			width = Math.min(Math.max(width, MIN_WIDTH), maxWidth);
+		}
+
+		return width;
 	}
 
 	render() {
 		this.logger.log("render");
-
```
tomlyn, line 30: Just a trivial thing but ... please don't remove lines like this used for spacing unless there is a very good reason to remove it.

canvas_modules/common-canvas/src/object-model/object-model.js
```diff
@@ -1298,6 +1298,10 @@ export default class ObjectModel {
 		this.store.dispatch({ type: "SET_BOTTOM_PANEL_CONFIG", data: { config: { panelHeight: ht } } });
 	}
 
+	setRightPanelWidth(wdth) {
+		this.store.dispatch({ type: "SET_RIGHT_FLYOUT_CONFIG", data: { config: { panelWidth: wdth } } });
+	}
+
```
tomlyn, line 1304: If you look at the comments this is in a section for 'Bottom panel methods'.  There is a section above for 'Right flyout methods' where this code should go.  

Also in the 'Right flyout methods'  can we add a `isRightFlyoutOpen()` method to keep things consistent?

tomlyn (changes requested):

tomlyn:
Well if it works I suppose it's OK but it isn't the same as the way bottom panel is structured which is what I was looking for or -- as I said previously -- if the way you have written the new code is somehow better than the way bottom panel is done then we should discuss making bottom panel work the same way. I probably haven't looked at it closely enough but I'm not sure why you are using the `ref`

Also, if we are going to use refs, I thought the appropriate way to create them is to use the `createRef()` function
https://react.dev/reference/react/createRef  At least that is how I have created them in other class components that we have in Common Canvas.

tomlyn:
Also, don't forget we will need a few Cypress tests for this new function.

srikant-ch5:
> Well if it works I suppose it's OK but it isn't the same as the way bottom panel is structured which is what I was looking for or -- as I said previously -- if the way you have written the new code is somehow better than the way bottom panel is done then we should discuss making bottom panel work the same way. I probably haven't looked at it closely enough but I'm not sure why you are using the `ref`

I have updated the code to be pretty much like `bottom-panel` apart from the `limitWidth` function because as per requirement panel should should extend upto `70%` leaving `30%` width for canvas. Unlike `bottom-panel` `right-panel` is a part of a grid.

> Also, if we are going to use refs, I thought the appropriate way to create them is to use the `createRef()` function https://react.dev/reference/react/createRef At least that is how I have created them in other class components that we have in Common Canvas.

I have removed `refs`. Thanks.

> Also, don't forget we will need a few Cypress tests for this new function.

I have included some Cypress Tests for the new function. In the build I can see the few Cypress Tests are failing which when I test in my local (even in local prod env) they are passing.

canvas_modules/common-canvas/src/common-canvas/cc-panels.jsx
srikant-ch5: Updated to include `isRightPanelOpen` also updated `common-canvas-test.js` as the condition `{ showRightFlyout: false }` was failing because of the change.

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
```diff
@@ -44,18 +44,43 @@ $panel-border-color: $border-subtle-01;
...
 }
 
+.right-flyout-drag {
+	cursor: ew-resize;
+	flex: 0 0 0.5px;
+	border: $panel-border-color;
+	background: $layer-01;
+	transition: all 0.2s ease-in;
+	&:hover {
+		flex: 0 0 $spacing-01;
```
tomlyn, line 79: What is this for - is it necessary?  I took it out and the right flyout seems to size OK. If this is needed to get Common Properties to display correctly? If so it ought to be part of the Common Properties CSS, not here.

tomlyn (changes requested):
@srikant-ch5 Looks good so far. I added a few comments.

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -589,21 +604,24 @@ class PropertiesMain extends React.Component {
...
+							ref={ (ref) => (this.commonProperties = ref) }
+							className={className}
+							onBlur={this.onBlur}
+							style={overrideStyle}
+						>
+							{propertiesTitle}
+							{propertiesDialog}
+							{buttonsContainer}
+						</aside>
+						{resizeBtn}
```
nmgokhale, line 623: Should we hide `resizeBtn` when `enableRightFlyoutDragToResize` is set?

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
srikant-ch5: I have updated above to include `spacing-01` in `right-flyout-drag`. 

But the problem with above spacing is that when a `common properties` `Accordion` is highlighted on click there is a small gap between right flyout left border and common properties.
Screenshot 2024-11-12 at 10 15 53 PM

I tried to resolve this using common properties css but this is visible in canvas so initially I set the width to `0.5px` so that the gap isn't visible and on hover it will take up `1px` width to match with bottom panel, but I have reverted it now and kept `spacing-01` for the `right-flyout-drag`.

https://github.com/user-attachments/assets/304382b9-5e7b-4cc4-ac8e-4faaaac47ddf

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -420,11 +427,19 @@ class PropertiesMain extends React.Component {
 		this.setState({ showPropertiesButtons: state });
 	}
 
+	updateRightFlyoutWidth(size) {
+		const element = document.querySelector(".common-canvas .right-flyout-container");
```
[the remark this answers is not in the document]
srikant-ch5: @nmgokhale Would appreciate your feedback on above I can see similar common properties dependancy on common canvas over [here](https://github.com/elyra-ai/canvas/blob/32f2a74d4d85844bb49e2f94b12b16c11aab81c8/canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.scss#L73) 

Without enforcing width to `right-flyout` in `properties` it will look something like below, where the properties content width is set as per custom classnames but flyout width is not reset.

Screenshot 2024-11-12 at 10 52 49 PM

canvas_modules/common-canvas/src/object-model/redux/canvas-store.js
```diff
@@ -75,7 +75,7 @@ export default class CanavasStore {
 			texttoolbar: { isOpen: false },
 			contextmenu: { isOpen: false, menuDef: [], source: {} },
 			leftflyout: {},
-			rightflyout: {},
+			rightflyout: { panelWidth: 300 },
```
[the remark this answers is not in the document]
tomlyn: The current right-flyout behavior allows an application to add some content into the right-flyout and it stretches to accommodate that content. (Remember applications can put whatever content they like into the right-flyout. They don't necessarily have to put Common Properties in there.)

So we need to make sure that function still works. Does that still work if we set the size to 0?

canvas_modules/common-canvas/src/common-canvas/common-canvas.scss
tomlyn: If you are concerned about the visual appearance of the UI after you've implemented a change you should be asking @Analuisa-Del-Rivero1  for her decision on how she wants it to appear. Unfortunately your vide doesn't show what it looks like with the accordion highlighting.

tomlyn: BTW It's probably best to ask question to Ana in the issue to keep all the design information discussion together in one place.

tomlyn: Perhaps you could do a video (or two?) that shows both approaches and what the accordion looks like in each case and explain to her about the difficulty in hovering over the drag area ... and see what she decides.

srikant-ch5:
> The resize button has some odd behaviors. I sometime see this after using the manual resize. https://github.com/user-attachments/assets/9945f6d2-4958-48fc-a17b-ecfe7e89d1e4

Thanks. I’ve resolved this by adopting a different approach to hide and unhide the button using a custom width array. Now, the behaviour is as follows: if the flyout is dragged, the button will be hidden. Once the flyout is returned to its smallest set width, the button will reappear. This approach provides the flexibility to restore the button without needing a refresh.


https://github.com/user-attachments/assets/64c1e819-e45c-477b-9d59-49403c934098


> Also, should the editor go back to the default size when opening a new panel? Right now it seems to go back to the last value.

As per confirmation from Ana it should behave just like `bottom panel`. Bottom panel currently goes back to the last set width like below.

https://github.com/user-attachments/assets/4964164f-176f-4967-9997-70be8dbc7422

canvas_modules/common-canvas/src/common-properties/properties-main/properties-main.jsx
```diff
@@ -38,7 +38,7 @@ import TitleEditor from "./../components/title-editor";
 import classNames from "classnames";
 
 import { injectIntl } from "react-intl";
-import styles from "./properties-main-widths.scss";
+import styles from "./properties-main-widths.module.scss";
```
matthoward366, line 41: Why rename?

matthoward366 (changes requested):
If I expand the window with the button then I can expand it any further with the resize bar.  

The resize bar should always work but the button can be hidden if the user uses the resize bar.

Also, if I open another property editor is there a way to reset it back to the default?  For example if I expand the editor using the resize bar to take up 1/2 the screen, close the editor, then open another editor, we should be able to set that to the default.

srikant-ch5:
> Also, if I open another property editor is there a way to reset it back to the default? For example if I expand the editor using the resize bar to take up 1/2 the screen, close the editor, then open another editor, we should be able to set that to the default.

To set it to default width I have included `componentWillUnmount` in `cc-right-flyout` to reset width. To match with bottom panel behaviour didn't include this earlier.
@tomlyn should we include this in `bottom-panel` too ?

matthoward366 (changes requested):
I can't resize editor when using pixel width.
Screenshot 2024-11-20 at 8 56 26 AM

srikant-ch5:
> I can't resize editor when using pixel width.

Yes, this was because of intital width in `cc-right-flyout`. This issue is fixed now, please have a look.

matthoward366 (approved):
lgtm

matthoward366:
@tomlyn can you have another look to make sure everything looks good?

tomlyn (approved):
LGTM

PR #2225, opened 2024-10-28
#2238 Write function to compare transform string in Cypress tests

Jerinjk14:
Fixes: #2238  

           in the file; or

           it.

### DIFF
--- a/canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
+++ b/canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
@@ -101,7 +101,6 @@ describe("Test zoomToReveal function returns the appropriate zoom object for a n
 		cy.verifyCanvasTransform(undefined);
 		// Zooming the canvas.
 		cy.clickToolbarZoomOut();
-		// TODO -- write function to test these +/- 2 px.
 		cy.verifyCanvasTransform("translate(45.81818181818181,32.22727272727275) scale(0.9090909090909091)");
 		// Test get Zoom to reveal & ZoomTo
 		cy.selectEntryFromDropdown("Histogram");
@@ -150,7 +149,6 @@ describe("Test zoomToReveal function returns the appropriate zoom object for a l
 		cy.verifyCanvasTransform(undefined);
 		// Zooming the canvas.
 		cy.clickToolbarZoomOut();
-		// TODO -- write function to test these +/- 2 px.
 		cy.verifyCanvasTransform("translate(45.81818181818181,32.22727272727275) scale(0.9090909090909091)");
 		// Test get Zoom to reveal & ZoomTo
 		cy.selectEntryFromDropdown("Binding (entry) node-Execution node");
@@ -158,7 +156,6 @@ describe("Test zoomToReveal function returns the appropriate zoom object for a l
 		cy.setYPercentOffset(60);
 		cy.submitAPI();
 		cy.wait(500);
-		// TODO -- write function to test these +/- 2 px.
 		cy.verifyCanvasTransform("translate(498.32727272727266,290.85454545454536) scale(0.9090909090909091)");
 	});
 });
--- a/canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
+++ b/canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
@@ -77,7 +77,6 @@ describe("Test zoomToReveal function returns the appropriate zoom object for a n
 		cy.setXPercentOffset(25);
 		cy.setYPercentOffset(25);
 		cy.submitAPI();
-		cy.wait(500);
 		cy.verifyCanvasTransform("translate(-1529.2531391552911,-684.6045237056301) scale(1)");
 	});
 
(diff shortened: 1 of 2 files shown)


canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
```diff
@@ -138,22 +141,22 @@ describe("Test zoomToReveal function returns the appropriate zoom object for a l
...
+		cy.verifyCanvasTransform("translate(45.81818181818181,32.22727272727275) scale(0.9090909090909091)");
 		// Test get Zoom to reveal & ZoomTo
 		cy.selectEntryFromDropdown("Binding (entry) node-Execution node");
 		cy.setXPercentOffset(70);
 		cy.setYPercentOffset(60);
 		cy.submitAPI();
-		// TODO -- write function to test these +/- 2 px.
-		// cy.verifyCanvasTransform("translate(498.32727272727266,290.85454545454536) scale(0.9090909090909091)");
+		cy.wait(500);
+		cy.verifyCanvasTransform("translate(498.32727272727266,290.85454545454536) scale(0.9090909090909091)");
```
tomlyn, line 159: Did you test to see if your code will return an error if an incorrect value is provided? I tried changing this line to:
```
cy.verifyCanvasTransform("translate(200,290.85454545454536) scale(0.9090909090909091)");
``` 
That is, with 200 as the x parameter and the test passed! 

When writing any tests, it is just as important to make sure the test will fail, in an invalid situation,  as it is to make sure the test passes.  If the test doesn't fail in the invalid situation then there would be no reason to run the test in the first place -- right?

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
```diff
@@ -1369,9 +1369,21 @@ Cypress.Commands.add("verifyNotificationIconType", (type) => {
 });
 
 Cypress.Commands.add("verifyCanvasTransform", (movString) => {
-	cy.get("#canvas-div-0 .d3-canvas-group")
-		.invoke("attr", "transform")
-		.should("eq", movString);
+	// Verify the argument passed is a string
+	if (movString === String) {
+		cy.get("#canvas-div-0 .d3-canvas-group", { timeout: 10000 })
```
tomlyn, line 1374: Is the `timeout` needed here. I see from the docs that the default is 4000 (4 seconds) which should be ample time for anything to happen in Common Canvas. We don't set this timeout anywhere else in the Cypress tests so if it isn't needed please remove it.

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
```diff
@@ -1369,9 +1369,21 @@ Cypress.Commands.add("verifyNotificationIconType", (type) => {
...
+			.then((actualTransformString) => {
+				const expectedValues = parseTransformString(movString);
+				const actualValues = parseTransformString(actualTransformString);
+
+				// Compare each part of the transform with compareRange
+				compareCloseTo(actualValues.translateX, expectedValues.translateX);
+				compareCloseTo(actualValues.translateY, expectedValues.translateY);
+				compareCloseTo(actualValues.scale, expectedValues.scale);
+			});
+	}
```
tomlyn, line 1386: I know it ought to be unlikely but, what happens here if `movString` is not a string? Will the function return as if the comparison succeeded? If so that not what we want as we want the test to fail in all error situations.

If you are testing for a string higher up in the function because sometimes zoom.cy.js passes in an undefined value then your code should specifically test for "undefined" and return success and in any other case (other than "string" ) return failure. In fact, since undefined is a value that is specifically passed in by zoom.cy.js, what the function ought to do is make sure the `actualTransformString` is undefined in that situation.

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
```diff
@@ -1369,9 +1369,21 @@ Cypress.Commands.add("verifyNotificationIconType", (type) => {
 });
 
 Cypress.Commands.add("verifyCanvasTransform", (movString) => {
-	cy.get("#canvas-div-0 .d3-canvas-group")
-		.invoke("attr", "transform")
-		.should("eq", movString);
+	// Verify the argument passed is a string
+	if (movString === String) {
```
tomlyn, line 1373: This isn't how you test for a string. This line of code is the reason invalid test situations are returning success (as mentioned in my comment above). You should use `typeof` to test for the type of a variable.

tomlyn (changes requested):

canvas_modules/harness/cypress/e2e/canvas/zoom.cy.js
Jerinjk14: I can see this error is occurring due to logic of checking the variable type, whether its a string or not, changed with 
if (typeof movString === "string") ,
now if the pixel range is other than +/-2 px, the test will fail
image

canvas_modules/harness/cypress/support/canvas/verification-cmds.js
Jerinjk14: I am able to fix this with if,else condition to check whether the movString is undefined, also if movString is not a string the test will fail with below error,
image

Jerinjk14:
@tomlyn ,I have pushed new commit as the issue with the cy.wait is because of using .then instead of .should,  we can ignore the unnecessary wait time now.

tomlyn (approved):
LGTM

PR #2335, opened 2025-02-06
#2334 Add new config option to link auto created node to selected binding node

### LINKED ISSUE #2334: New config option to allow auto creating a node linked to a selected binding node
ref planning 17075

Currently the auto node creation does not automatically link to a binding node. It will look for the next available node within the branch that has available output ports to link with. This new config option will allow the newly created node to be linked with a terminal node as an input.

caritaou:
closes #2334

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/object-model/api-pipeline.js
+++ b/canvas_modules/common-canvas/src/object-model/api-pipeline.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 Elyra Authors
+ * Copyright 2017-2025 Elyra Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -320,61 +320,72 @@ export default class APIPipeline {
...
+		}
+		return viableNode;
 	}
 
 	// Returns a newly created 'auto node' whose position is based on the
-	// source node (if one is provided) and the the other nodes on the canvas.
-	createAutoNode(data, sourceNode) {
+	// selected node (if one is provided) and the the other nodes on the canvas.
+	// If autoLinkToBindingNodes is true, then position the new node to the left of the selected node
+	createAutoNode(data, selectedNode, autoLinkToBindingNodes) {
```
tomlyn, line 370: This isn't the necessarily the selected node -- from the caller it is the 'nodeToLinkWith` so you could call it the same here.

canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -320,61 +320,72 @@ export default class APIPipeline {
...
 		const initialMarginX = this.objectModel.getCanvasLayout().autoLayoutInitialMarginX;
 		const initialMarginY = this.objectModel.getCanvasLayout().autoLayoutInitialMarginY;
 		const horizontalSpacing = this.objectModel.getCanvasLayout().autoLayoutHorizontalSpacing;
 		const verticalSpacing = this.objectModel.getCanvasLayout().autoLayoutVerticalSpacing;
+		const isSelectedNodeBinding = selectedNode ? this.isExitBindingNode(selectedNode) : false;
```
tomlyn, line 375: I would do this test outside of this function (in the calling code) and from the calling code decide if I want the new node placed on the left or right of the  'nodeToLinkWith'. Then pass in a boolean to this method instead of `autoLinkToBindingNodes` which would indicate position is either to the left or right of 'nodeToLinkWith'.  That way you won't have to have `autoLinkToBindingNodes && isSelectedNodeBinding`  twice in the code below and this method becomes a simple utility that decides where to put the new node rather than understanding about exit binding nodes  and config values etc

canvas_modules/common-canvas/src/object-model/api-pipeline.js
```diff
@@ -320,61 +320,72 @@ export default class APIPipeline {
...
+	// Returns a node for auto completion or null if no node can be
+	// detected. The node is either:
 	// 1. The selected node, if only *one* node is currently selected or
 	// 2. The most recently added node, provided it has one or more output ports or
 	// 3. The most-recent-but-one added node, provided it has one or more output ports
-	getAutoSourceNode(autoLinkOnlyFromSelNodes) {
-		var sourceNode = null;
+	//    or if autoLinkToBindingNodes is true, then it may be a binding node
+	getAutoLinkedNode(autoLinkOnlyFromSelNodes, autoLinkToBindingNodes) {
+		var autoLinkedNode = null;
```
tomlyn, line 330: Instead if using `autoLinkedNode` why not use the name the caller calls it when it is returned which is `nodeToLinkWith`. That way, we are using the same name for the same concept.

tomlyn (changes requested):

canvas_modules/common-canvas/src/object-model/api-pipeline.js
caritaou: Good point, updated

caritaou:
This is no longer needed

Issue #2545, opened 2025-05-14
Clicking the tooltip of an action button in the toolbar triggers the action


srikant-ch5:
Ref: 4075

PR #2546, opened 2025-05-14
#2545 Clicking the tooltip of an action button in the toolbar triggers the action

srikant-ch5:
Fixes: #2545
 
           in the file; or

           it.


tomlyn:
Hi @srikant-ch5 

Thanks for having a go at fixing this. I am not sure this is the best way to do it for 2 reasons:
1. It will still allow the background of 'hoverable' tooltips to be clicked which will then perform the toolbar button action. I'm not sure the user will expect that. I have asked a question here about it so let's see what Jennifer says: https://ibm-analytics.slack.com/archives/C598XEX8S/p1747251164346469
2. This PR would change the behavior of the toolbar API because, with this change, applications would have to specify the `hoverable` property for tooltips where they want that behavior. As you know, if we change the API behavior (other than adding new stuff) we should do that in a major release not a minor or patch release because in those types of releases  applications should not have to make code changes to get the same behavior.

I think the solution might be to catch the `onMousedown` and `onClick` in the tooltip and stop propagation.

tomlyn (changes requested):
See my comment in the conversation. I think we should fix this by suppressing the propagation when the user clicks on the background of the tooltip.

srikant-ch5:
> See my comment in the conversation. I think we should fix this by suppressing the propagation when the user clicks on the background of the tooltip.

Sure, I will update the PR accordingly.

srikant-ch5:
@tomlyn I have added function to suppress event propagation.

tomlyn (approved):
LGTM

Issue #2600, opened 2025-06-13
Sub-panel should act as a modal dialog (keyboard accessibility)


Jerinjk14:
ref: #4069

PR #2601, opened 2025-06-13
#2600 Sub-panel should act as a modal dialog (keyboard accessibility)…

Jerinjk14:
FIxes: #2600 
 
           in the file; or

           it.


matthoward366:
@nmgokhale can you review this?

canvas_modules/common-canvas/src/common-properties/components/wide-flyout/wide-flyout.jsx
```diff
@@ -66,6 +67,36 @@ export default class WideFlyout extends Component {
...
+			return;
+		}
+		const modal = document.querySelector(".properties-wf-content.show");
+		if (!modal) {
+			return;
+		}
+		// Always fetch focusables for latest DOM state.
+		this.focusables = Array.from(
+			modal.querySelectorAll(
+				"button, a[href], input, select, textarea, [tabindex]:not([tabindex='-1'])"
```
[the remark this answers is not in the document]
Jerinjk14: As per the requirement, need to trap the focus inside the modal, i.e tab/ shift+tab should works as expected inside the modal so we can query the current content as **const modal = document.querySelector(".properties-wf-content.show");**

Jerinjk14:
@matthoward366 , Yes we can either trap the focus using [focus-trap-react] or by our own trap method, I preferred handling ourselves instead of using the package as per your comment on the issue, If we need to use the focus-trap then we can go ahead with that, What's your suggestion about this?

Jerinjk14:
@nmgokhale Working as expected now, PFA

https://github.com/user-attachments/assets/46dccf52-36db-4dd8-ad64-6fedbc4e0fa8

nmgokhale:
@Jerinjk14 Thanks for fixing the issues I reported earlier. 

Earlier, the first element in modal used to come in focus only after I pressed `tab` key. Now after opening the modal, first element always shows focus even though I didn't press any `tab` key. @matthoward366 Is this behavior okay? 

https://github.com/user-attachments/assets/4c1a95cc-f19d-4ad6-872f-6a43eddd7252

matthoward366:
Usually you do want an element to have focus.  The hard part is figuring out which element that should be.  From what I can tell this seems fine.

canvas_modules/common-canvas/src/common-properties/components/wide-flyout/wide-flyout.jsx
```diff
@@ -24,23 +24,58 @@ import { Portal } from "react-portal";
...
+	componentDidUpdate(prevProps) {
+		// If modal is still open, and new item added.
+		const modal = this.modalRef.current;
+		if (!prevProps.show && this.props.show && modal) {
+			// If focus is outside modal, move it to first focusable
+			this.focusOnFirstFocusable();
+			return;
+		}
+		// Restore focus if lost due to modal content changes like typing or adding new fields
+		if (this.props.show && modal && document.activeElement === document.body) {
```
nmgokhale, line 53: When I close the modal after updating value, focus is no longer in the modal -

https://github.com/user-attachments/assets/c892126e-fdad-4f2f-b591-46060063638b

Jerinjk14: @nmgokhale This will fix:
- Focus will Trap Inside modal while mounting the modal, and focus will be on first element inside modal by default,
- If more than one wideflyout then the focus will be shifted to the second/outer modal and initial focus will be on first element- Tab will work as expected inside modal.
- Also while editing/modifying the existing value inside the table list focus wont be lost.
- Once save the edited. field focus will Trap inside, here initial focus wont be set, If you press Tab focus automatically trap inside modal on the first focusable element and if Shift+tab then focus will be on last element inside modal, and wont jumps out from modal.

nmgokhale (approved):

PR #2650, opened 2025-07-02
#2649 Table toolbar looks odd for Cancel Text - small flyout

### LINKED ISSUE #2649: Table toolbar looks odd for Cancel Text in Small flyout
Ref: 4099

Image

veenas1:
Fixes #2649
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/components/table-toolbar/table-toolbar.scss
+++ b/canvas_modules/common-canvas/src/common-properties/components/table-toolbar/table-toolbar.scss
@@ -25,6 +25,10 @@
 		@include type-style("label-02");
 		padding: 0 $spacing-05;
 		color: $text-on-color;
+		&.small-flyout{
+			padding: 0 $spacing-03;
+			min-width: 0;
+		}
 	}
 
 	.properties-action-list {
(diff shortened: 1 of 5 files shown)


canvas_modules/common-canvas/src/common-properties/components/table-toolbar/table-toolbar.scss
```diff
@@ -25,6 +25,10 @@
 		@include type-style("label-02");
 		padding: 0 $spacing-05;
 		color: $text-on-color;
+		&.small-flyout{
+			padding: 0 $spacing-03;
```
caritaou, line 29: You can remove the `small-flyout` class and padding. Setting `min-width: 0` should be sufficient in all cases. The text should be aligned with the checkbox
table-toolbar

veenas1: But if we do not adjust the padding, the end of text(`items`) gets hidden.
Also, when we expand and narrow down the panel, the `props.smallFlyout` is not getting updated, only on clicking the checkbox, there is update and so the class reflects.

image

caritaou: This will also be an issue for other languages, so adding this small "fix" for English is not necessary IMO. I think this will be addressed once we moved to use the canvas toolbar here

veenas1: updated the PR to have only the min-width change.
 Remaining can be done as part of moving to canvas-toolbar.

caritaou (approved):

nmgokhale (approved):

Issue #2674, opened 2025-07-25
Enhancement : Support for icons in action buttons


Jerinjk14:
ref: #3953

PR #2675, opened 2025-07-25
#2674 Enhancement : Support for icons in action buttons-implementing

Jerinjk14:
Fixes: #2674 
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/carbon.scss
+++ b/canvas_modules/common-canvas/src/carbon.scss
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Elyra Authors
+ * Copyright 2024-2025 Elyra Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,3 +20,4 @@
 @forward "@carbon/react/scss/theme";
 @forward "@carbon/react/scss/motion";
 @forward "@carbon/react/scss/components/button/tokens";
+@forward "@carbon/react/scss/components/tag/tokens";
--- a/canvas_modules/common-canvas/src/palette/palette.scss
+++ b/canvas_modules/common-canvas/src/palette/palette.scss
@@ -134,8 +134,8 @@ $palette-dialog-list-item-height: 46px;
 			padding: carbon.$spacing-05 10px 0 58px;
 
 			& mark {
-				color: carbon.$interactive;
-				background-color: carbon.$highlight;
+				color: carbon.$tag-color-blue;
+				background-color: carbon.$tag-background-blue;
 				font-weight: 500;
 			}
 		}
@@ -161,8 +161,8 @@ $palette-dialog-list-item-height: 46px;
 			width: calc(100% - 68px);
 			line-height: 1;
 			& mark {
-				color: carbon.$button-primary-active;
-				background-color: carbon.$highlight;
+				color: carbon.$tag-color-blue;
+				background-color: carbon.$tag-background-blue;
 				font-weight: 500;
 			}
 		}
@@ -182,8 +182,8 @@ $palette-dialog-list-item-height: 46px;
 			width: 100%; // Needed when the description is only one line long
 
 			& mark {
-				color: carbon.$button-primary-active;
-				background-color: carbon.$highlight;
+				color: carbon.$tag-color-blue;
+				background-color: carbon.$tag-background-blue;
 				font-weight: 500;
 			}
 		}
--- a/canvas_modules/harness/assets/styles/carbon.scss
+++ b/canvas_modules/harness/assets/styles/carbon.scss
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Elyra Authors
+ * Copyright 2024-2025 Elyra Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-@forward "@carbon/react/scss/colors";
-@forward "@carbon/react/scss/type";
-@forward "@carbon/react/scss/spacing";
-@forward "@carbon/react/scss/theme";
-@forward "@carbon/react/scss/motion";
-@forward "@carbon/react/scss/components/button/tokens";
+@forward "@carbon/react" with (
+	$font-path: "/fonts/plex",
+	$use-per-family-plex: true
+);
+
+
--- a/canvas_modules/harness/assets/styles/common.scss
+++ b/canvas_modules/harness/assets/styles/common.scss
@@ -13,10 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@use "@carbon/react" with (
-	$font-path: "/fonts/plex",
-	$use-per-family-plex: true
-);
 
 @use "@ibm/plex-sans-condensed/scss" as PlexSansCondensed with (
 	$font-prefix: "/fonts/plex-sans-condensed"
--- a/canvas_modules/harness/assets/styles/harness.scss
+++ b/canvas_modules/harness/assets/styles/harness.scss
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-@use "@carbon/react/scss/themes";
+
 @use "./carbon.scss";
+@use "@carbon/react/scss/themes";
 
 @forward "../../src/client/components/console/console.scss";
 @forward "../../src/styles/custom-canvases.scss";
--- a/canvas_modules/harness/src/client/components/custom-canvases/all-ports/all-ports.scss
+++ b/canvas_modules/harness/src/client/components/custom-canvases/all-ports/all-ports.scss
@@ -41,7 +41,7 @@
 
 	// Style new connection link line
 	.d3-new-connection-line[linkType="nodeLink"] {
-		stroke: carbon.$button-primary;
+		stroke: carbon.$interactive;
 		stroke-width: 2;
 		stroke-dasharray: 1 0;
 		fill: none;
--- a/canvas_modules/harness/src/client/components/custom-canvases/flows/flows.scss
+++ b/canvas_modules/harness/src/client/components/custom-canvases/flows/flows.scss
@@ -30,7 +30,7 @@
 
 		&[data-new-link-over="yes"] {
 			.d3-node-selection-highlight {
-				stroke: carbon.$button-primary;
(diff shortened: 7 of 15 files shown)


canvas_modules/common-canvas/__tests__/common-properties/actions/button-action-test.js
```diff
@@ -244,5 +244,8 @@ describe("actions using paramDef", () => {
 		// class_name not defined in uiHints action_info
 		const decrementButton = container.querySelector("div[data-id='decrement']");
 		expect(decrementButton.className).to.equal("properties-action-button");
+
+		const iconButton = container.querySelector("div[data-id='iconButton']");
```
matthoward366, line 248: Shouldn't there be a test to validate callback is called and the icon is displayed?

canvas_modules/common-canvas/src/common-properties/actions/action-factory.js
```diff
@@ -28,18 +28,49 @@ export default class ActionFactory {
...
 			if (action.actionType === ActionType.BUTTON) {
-				return (
-					<ButtonAction
-						key={"action." + key}
-						action={action}
-						controller={this.controller}
-					/>
-				);
+				let actionButton;
+				if (action.name === "iconButton") {
```
matthoward366, line 54: This seems wrong.  Wouldn't this require all action names to be iconButton?

canvas_modules/common-canvas/src/common-properties/actions/action-factory.js
```diff
@@ -28,18 +28,49 @@ export default class ActionFactory {
...
+					actionButton = (
+						<ButtonAction
```
matthoward366, line 56: I don't understand why this extra code is needed.  All this can be done in ButtonAction without having to pass the extra parameters.

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -75,6 +75,8 @@ class ButtonAction extends React.Component {
 		const disabled = this.props.state === STATES.DISABLED;
 		const actionButtonKind = this.getActionButtonKind();
 		const actionButtonSize = this.getActionButtonSize();
+		// Get the icon to be rendered from the call back
+		const icon = this.props.buttonIconHandler?.(this.props.action.name, this.props.action);
```
matthoward366, line 79: This doesn't need to be a new prop.  You can get the handler from the controller.

canvas_modules/harness/src/client/App.js
```diff
@@ -1446,6 +1446,16 @@ class App extends React.Component {
 		// handle custom buttons icon
 		if (data.type === "customButtonIcon") {
 			callbackIcon(<Edit size={32} />);
+		} else if (data.type === "actionButtonIcon") { // handle icons for action button
+			console.log("data",data);
+			const iconName = data.data.data.icon;
+			const iconComponenet = require("@carbon/icons-react")[iconName];
```
matthoward366, line 1452: This doesn't have to be dynamic.  Just hardcode an icon based on the actionId.

canvas_modules/harness/src/client/components/common-properties-components.jsx
```diff
@@ -227,6 +233,12 @@ class CommonPropertiesComponents extends React.Component {
 			let value = this.flyoutController.getPropertyValue(propertyId);
 			this.flyoutController.updatePropertyValue(propertyId, value -= 1);
 		}
+		if (actionId === "iconButton") {
```
matthoward366, line 236: Why do have if this same code 3 times to update the same property value?

matthoward366 (changes requested):

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -83,6 +104,8 @@ class ButtonAction extends React.Component {
 				onClick={this.applyAction}
 				disabled={disabled}
 				title={this.props.action.label.text}
+				renderIcon={icon || null}
+				iconDescription={ icon ? (this.props.action?.description?.text || "Action Button icon") : null } // Text to appear in Tooltip
```
matthoward366, line 108: `Action Button icon` should be translated and sentence case.

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -27,10 +27,29 @@ class ButtonAction extends React.Component {
...
+		const buttonIconHandler = controller.getHandlers().buttonIconHandler;
+		if (buttonIconHandler && action) {
+			const propertyId = { name: action.data.parameter_ref };
+			buttonIconHandler({
+				type: "actionButtonIcon",
+				propertyId: propertyId,
+				buttonId: action.name,
+				data: action
+			}, (appIcon) => {
+				this.setState({ icon: appIcon });
```
matthoward366, line 47: Why not do this in the constructer then we wouldn't need to set the icon in the state.  That would also remove the case where `null` is nothing is returned.

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -27,10 +27,29 @@ class ButtonAction extends React.Component {
...
 	}
 
+	// Load icon via buttonIconHandler with a new type "actionButtonIcon" after mount
+	componentDidMount() {
+		const { controller, action } = this.props;
+		const buttonIconHandler = controller.getHandlers().buttonIconHandler;
+		if (buttonIconHandler && action) {
+			const propertyId = { name: action.data.parameter_ref };
+			buttonIconHandler({
+				type: "actionButtonIcon",
```
matthoward366, line 42: I think this should probably be a constant.

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -27,10 +27,29 @@ class ButtonAction extends React.Component {
...
 		};
 		this.applyAction = this.applyAction.bind(this);
 	}
 
+	// Load icon via buttonIconHandler with a new type "actionButtonIcon" after mount
+	componentDidMount() {
+		const { controller, action } = this.props;
+		const buttonIconHandler = controller.getHandlers().buttonIconHandler;
+		if (buttonIconHandler && action) {
+			const propertyId = { name: action.data.parameter_ref };
```
matthoward366, line 40: What is action.data isn't defined?  I would probably call this similar to 
```
actionHandler(this.props.action.name,
				this.props.controller.getAppData(), this.props.action.data)
```

matthoward366: I don't think we'd pass `propertyId`.   I would think these are the values we'd pass in this case.  We'd have to make sure the documentation and typeScript is updated correctly for this.
```
const iconData = {
type: "actionButtonIcon",
buttonId: action.name,
data: action.data
};
```

canvas_modules/common-canvas/__tests__/common-properties/actions/button-action-test.js
```diff
@@ -17,19 +17,23 @@
 import React from "react";
 import { Provider } from "react-redux";
 import ActionButton from "./../../../src/common-properties/actions/button";
-import { render } from "../../_utils_/mount-utils.js";
 import { expect } from "chai";
-import { expect as expectJest } from "@jest/globals";
+import { afterEach, beforeEach, expect as expectJest } from "@jest/globals";
```
matthoward366, line 21: Why did you need to add `afterEach` and `beforeEach`?  These should already be globally defined.

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -22,12 +22,27 @@ import { STATES, CARBON_BUTTON_KIND, CARBON_BUTTON_SIZE } from "./../../constant
...
-		this.state = {
-		};
+		this.icon = null;
+		const { controller, action } = this.props;
+		const buttonIconHandler = controller.getHandlers().buttonIconHandler;
+		if (buttonIconHandler && action) {
+			const iconData = {
+				type: "actionButtonIcon",
+				buttonId: action.name,
+				data: action
```
matthoward366, line 38: Shouldn't this be action.data?

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
```diff
@@ -83,6 +104,8 @@ class ButtonAction extends React.Component {
 				onClick={this.applyAction}
 				disabled={disabled}
 				title={this.props.action.label.text}
+				renderIcon={this.icon}
```
matthoward366, line 107: Instead of passing in null here in the cases where there is no icon this might be better to defined those as an object above and use the ... to add the properties if they are defined.  

Something like
```
const iconProps = {};
if (this.icon) {
  iconProps.renderIcon=this.icon
  iiconProps.conDescription= iconDescription
  }

File: docs/pages/04.02-callbacks.md [hunk 14]
```diff
@@ -99,8 +99,11 @@ buttonHandler(data) {
 }
 ```
 
-### buttonIconHandler()
-Called when there is a `buttons` uihints set in the `complex_type_info` section of the parameter definition. This buttonIconHandler expects a Carbon Icon jsx object as the return value from the callback. This is used to display the Carbon icon in the custom table button. The buttonIconHandler provides the following data:
+### buttonIconHandler
+ This handler is used to display the Carbon icon in the custom table button and action buttons based on the `type` field of the data. This buttonIconHandler expects a Carbon Icon jsx object as the return value from the callback. The buttonIconHandler provides the following `type`:
```
matthoward366, line 103: This doesn't have to be a carbon icon.  It's just a react object.

matthoward366 (commented):
Getting close.  A couple minor comments.

docs/pages/04.02-callbacks.md
Jerinjk14: This handler is used to display the Carbon icon in the custom table button and action buttons based on the `type` field of the data. _The buttonIconHandler expects a React object to be returned via the provided `callbackIcon` function_. The buttonIconHandler provides the following `type`:
**This seems fine?**

canvas_modules/common-canvas/src/common-properties/actions/button/button.jsx
Jerinjk14: Sure, made changes with the spread

Issue #2700, opened 2025-08-22
callbackIcon parameter needs to be fixed


BonbonLemon:
In common-properties.ts

export type ButtonIconHandler = (
  data: CustomButtonIconData | ActionButtonIconData,
  callbackIcon: (icon: ReactNode) => void
) => void;

callbackIcon is expecting a ReactNode, but it gives me an error when I try to pass <Launch />
However, when I pass in Launch (just the CarbonIconType), then it works.

My workaround for this (working with strict typing) is that I'm casting Launch as unknown and then as ReactNode 
callbackIcon(Launch as unknown as ReactNode).

Either needs to fix the typing expected, or change what callbackIcon is doing.

PR #2722, opened 2025-09-22
callbackIcon parameter needs to be fixed-fixing_component-type-to-jsx…

Jerinjk14:
Fixes : #2700 

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
+++ b/canvas_modules/common-canvas/src/common-canvas/canvas-controller.js
@@ -1741,6 +1741,10 @@ export default class CanvasController {
 		this.objectModel.setRightFlyoutWidth(wd);
 	}
 
+	setRightFlyoutMinWidth(wd) {
+		this.objectModel.setRightFlyoutMinWidth(wd);
+	}
+
 	isRightFlyoutOpen() {
 		return this.objectModel.isRightFlyoutOpen();
 	}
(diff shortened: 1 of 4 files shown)


Jerinjk14:
@matthoward366 Carbon's renderIcon prop will expect a `component type ` instead of reactNode, if we also need to pass `Launch` instead of `<Launch />`, then we can change type like below
```
export type ButtonIconHandler = (
  data: CustomButtonIconData | ActionButtonIconData,
  callbackIcon: (icon: ReactNode | ComponentType<any> | null) => void
) => void;
```
Then we can use both component type(Launch) and JSX (Launch), let me know if we can stay with reactNode itself or need to implement component type as well

nmgokhale (approved):

canvas_modules/harness/src/client/App.js
```diff
@@ -1457,13 +1457,13 @@ class App extends React.Component {
 			const actionId = data.buttonId;
 			switch (actionId) {
 			case "iconButton":
-				iconComponenet = Launch;
+				iconComponenet = <Launch />;
 				break;
 			case "increment1":
-				iconComponenet = Add;
+				iconComponenet = <Add />;
```
nmgokhale, line 1463: @Jerinjk14 Can you remove `< />` from one of the icons to make sure both `Add` and `<Launch />` work okay?

Jerinjk14: Sure, changed `<Launch /> `to its component type `Launch`. working as expected

PR #3100, opened 2026-04-16
3099 Ensure NPM packages being used do not have vulnerabilities

### LINKED ISSUE #3099: Ensure NPM packages being used do not have vulnerabilities
Ref: 4349

tomlyn:
Fixes: #3099 
 
           in the file; or

           it.


tomlyn:
@matthoward366

 I ran the `npm audit fix --force` as suggested. There were a couple of issues:
 
1. There seems to be a circular problem in that it wants to install `grunt-jsonlint` v3.0.0 but after that is installed it finds that v3.0.0 has a dependency problem so it wants to install v2.1.5 instead but that then has a problem so it wants to go back to v3.0.0. I ended up just setting it to v2.1.5. Maybe there will be a fix in v3.0.0 but that package hasn't been published for 2 years so I'm not sure.
2. The `npm audit fix` wants to install `"stylelint": "17.8.0"` but after v15.0.0 there seems to have been a change in how the lint rules are handled which causes a lot of build errors. Bob suggested a way forward for that would be to use `Prettier` for formatting and `stylelint` for error checking - I'm not sure if that is the right way to go. Anyway, I ended up just setting `"stylelint"` to `v14.16.1` instead which is the last release before v15.0.0.

matthoward366 (approved):
This is fine.  In general build dependencies don't need to be updated.  Eventually we can look at moving away from grunt and updating the linting to the latest versions.  That should fix the majority of the issues.

PR #3201, opened 2026-06-30
#3197 Toolbar settings dropdown menu is cut off when canvas is zoomed in

### LINKED ISSUE #3197: Toolbar settings dropdown menu is cut off when canvas is zoomed in
Ref: 4380

nathayu:
… div

Fixes: #3197 

 
           in the file; or

           it.


michaelcheungkm:
LGTM reviewed the changes made again

tomlyn:
I think this needs more work. Here's a video showing what I noticed. Unzip and when you run it turn the sound on and volume up.

https://github.com/user-attachments/assets/b7817fc8-6514-431d-8ec7-7b770164641d

nathayu:
Fixed have been implemented
image

JesusGuerrero:
@tomlyn can you please review

canvas_modules/common-canvas/src/toolbar/toolbar-sub-utils.js
```diff
@@ -59,21 +117,42 @@ export function adjustSubAreaPosition(subAreaRef, containingDivId, expandDirecti
...
+
+		// If the cascade is itself taller than the available space, clamp and scroll.
+		const availableHeight = bounds.bottom - topPos;
+		if (menuHeight > availableHeight) {
+			setStyle(subAreaRef, "maxHeight", Math.max(availableHeight, 0) + "px");
+			setStyle(subAreaRef, "overflowY", "auto");
+		} else {
+			setStyle(subAreaRef, "maxHeight", "");
+			setStyle(subAreaRef, "overflowY", "");
 		}
```
tomlyn, line 143: Something in these changes seems to have messed up the horizontal positioning of the notification panel (see below).

Image

Here's what it looks like without this PR:

Image

 It should be being positioned relative to the button that opens it, which is the right most button in the toolbar. Typically a panel should open where its top left corner and the button's bottom left corner are adjacent.  However, in this case, that would partly obscure the panel because part of it would be outside the viewport. So, in this case, the positioning algorithm ensures the full panel is visible by moving it to the left far enough to make it fully visible inside the viewport.  This is the behavior we need to preserve.

Note: If you test with the "Sub Areas" test toolbar it is a little confusing because there are two notification panels available from that toolbar. One is displayed by clicking the right most toolbar icon (and is shown in the screen shots above). This is the current notification panel we ship with Common Canvas.

The other one appears when you click the 'Bell' icon and its contents are provided by the 'Carbon for IBM Products' [notification component](https://pages.github.ibm.com/carbon/ibm-products/components/notification-panel/usage/). This is just there as a proof of concept to show how an application development team could use that component in preference to the one provided by Common Canvas, if they want. 

Image

You can see that this also get pushed to the left by the positioning code if the Bell icon gets close to the right side of the viewport,  like this:

Image

canvas_modules/common-canvas/__tests__/toolbar/toolbar-sub-utils-test.js
```diff
@@ -0,0 +1,271 @@
+/*
+ * Copyright 2024 Elyra Authors
```
tomlyn: The copyright year number of newly created files should be the current year.

tomlyn (changes requested):

canvas_modules/common-canvas/src/toolbar/toolbar-sub-utils.js
```diff
@@ -83,15 +149,20 @@ export function generateSubAreaStyle(expandDirection, actionItemRect) {
 		return null;
 	}
 
+	// Cascade sub-areas use position:fixed so they are not clipped by a scrollable
+	// parent menu's overflow. Because they are fixed, the initial position is
+	// expressed in viewport coordinates (adjustSubAreaPosition refines it).
 	if (expandDirection === "vertical") {
 		return {
-			top: actionItemRect.height + 1,
+			position: "fixed",
```
tomlyn, line 157: Sorry I didn't notice this before but using `position: "fixed"` causes a minor issue. This setting ties the position of the object to the page's viewport which means that, if an application displays Common Canvas on a scrollable page and the cascaded sub-menu or sub-panel is open, if the user scrolls the page the sub-area (menu or panel) would remain fixed on the page as Common Canvas moves up or down. I know its an unlikely scenario but I'd rather not build that regression in to this fix if we can help it. 

I asked Bob about this and he said using `position: "fixed"` is probably the best approach so I asked him to add a scroll listener to detect when the page is being scrolled and close the cascaded sub-area when that happens. You can ask him to do the same or if not here is the file he generated for me. 

[toolbar-sub-menu-item.jsx.zip](https://github.com/user-attachments/files/29860176/toolbar-sub-menu-item.jsx.zip)

tomlyn (changes requested):

tomlyn (approved):
LGTM

PR #3215, opened 2026-07-08
#4376 Set row_selection to 'none' on Expression Builder Custom Value table

### LINKED ISSUE #3216: Set row_selection to 'none' on Expression Builder Custom Value table
ref: 4376

rbeshay1:
Fixes [#3216](https://github.com/elyra-ai/canvas/issues/3216)
 

           in the file; or

           it.


matthoward366:
Does this also allow adding values using keyboard navigation?

rbeshay1:
> Does this also allow adding values using keyboard navigation?

Yes, adding values via keyboard is still supported. Users can Tab to the + button in each row and press Enter to add a value or double click a row. The only thing removed is the ability to arrow key select a row, which was not functional here since selecting a value row had no effect without clicking add.

caritaou:
> The only thing removed is the ability to arrow key select a row, which was not functional here since selecting a value row had no effect without clicking add.

@rbeshay1 The table on the left requires row selection because that is what triggers the right table to render. The rows on the right table is dependent on what row is clicked from the left table

rbeshay1:
> > The only thing removed is the ability to arrow key select a row, which was not functional here since selecting a value row had no effect without clicking add.
> 
> @rbeshay1 The table on the left requires row selection because that is what triggers the right table to render. The rows on the right table is dependent on what row is clicked from the left table

@caritaou  I see. I was only referring to the right/values table there.

I agree the left fields table should continue using ROW_SELECTION.SINGLE, since selecting a field row is what populates/renders the values table on the right. My comment about removing row selection only applies to the right values table, where selecting a row by itself did not have any effect without clicking add.

caritaou (approved):
lgtm

PR #3247, opened 2026-07-21
#3246 Prepare for unsafe-inline - webpack extract & harness server

### LINKED ISSUE #3246: Prepare for unsafe-inline - webpack extract & harness server

tomlyn:
Fixes: #3246

Switch harness dev webpack from style-loader to MiniCssExtractPluginstyle-loader injects CSS by creating <style> tags at runtime via styleTagTransform.js / injectStylesIntoStyleTag.js. This requires 'unsafe-inline' in the CSP style-src directive.

MiniCssExtractPlugin serves styles as real .css files loaded via <link> tags, which are permitted under style-src 'self'. This allows the harness to run without 'unsafe-inline' in style-src.

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/__tests__/_utils_/custom-actions/CustomOverflowAction.js
+++ b/canvas_modules/common-canvas/__tests__/_utils_/custom-actions/CustomOverflowAction.js
@@ -34,16 +34,17 @@ class CustomOverflowAction {
 
 	renderAction() {
 		return (
-			<OverflowMenu
-				data-floating-menu-container
-				flipped={this.data?.parameter_ref === "oneofselect-custom-action-right"}
-				aria-label="Overflow menu"
-				size="sm"
-				className="harness-custom-action"
-			>
-				<OverflowMenuItem className="overflow-menu-item" itemText="Menu item 1" onClick={() => this.changeReadonlyText("Menu item 1")} />
-				<OverflowMenuItem className="overflow-menu-item" itemText="Menu item 2" onClick={() => this.changeReadonlyText("Menu item 2")} />
-			</OverflowMenu>
+			<div data-floating-menu-container style={{ position: "relative" }}>
+				<OverflowMenu
+					flipped={this.data?.parameter_ref === "oneofselect-custom-action-right"}
+					aria-label="Overflow menu"
+					size="sm"
+					className="harness-custom-action"
+				>
+					<OverflowMenuItem className="overflow-menu-item" itemText="Menu item 1" onClick={() => this.changeReadonlyText("Menu item 1")} />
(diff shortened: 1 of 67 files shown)


tomlyn:
@matthoward366 
I've been working on this issue [4374](https://github.ibm.com/NGP-TWC/wdp-abstract-canvas/issues/4374) which is a request to be able to run with CSP without needing to specify `style-src 'unsafe-inline'`.

I changed the code to remove the inline styles in separate PRs which are referenced in [my comment](https://github.ibm.com/NGP-TWC/wdp-abstract-canvas/issues/4374#issuecomment-223986995) in the issue. This PR is the final piece of the puzzle.

Here's Bob's explanation of what we are doing in this PR:

## Explanation - webpack.config.dev.cjs

This PR is the development infrastructure counterpart to all the source code changes — here's how it fits in:

### The problem with style-loader

In development mode, the harness webpack config was using style-loader. This works by injecting CSS at runtime: it creates `<style>` tags in the document <head> and writes the CSS content into them via JavaScript. Internally this uses element.innerHTML or similar DOM manipulation — which the browser treats as an inline style, requiring 'unsafe-inline' in the style-src CSP directive.

### What the source code changes achieved

All the PRs we've done so far — csp-toolbar, csp-color-picker, csp-floating-toolbars-context-menu, csp-layout-panels, csp-svg-canvas, csp-common-properties — each eliminated inline style attributes on DOM elements by moving values into CSS custom properties (set via CSSOM setProperty(), which CSP doesn't block). This gets rid of the need for 'unsafe-inline' in the page's per-element content.

### Why style-loader undoes all of that

Even with every inline style attribute gone from the React components, if the webpack dev bundle is still using style-loader, the browser will block the injected <style> tags for the same CSP reason. The stylesheet content itself would be blocked, making the entire app unstyled. So removing 'unsafe-inline' from the CSP would break the harness in dev mode regardless of all the component-level work.

### What MiniCssExtractPlugin does instead

MiniCssExtractPlugin extracts all the CSS into real .css files that are served as static assets and loaded via <link rel="stylesheet"> tags. The browser fetches them as external resources from the same origin, which is permitted under `style-src 'self'` — no 'unsafe-inline' needed.

## Explanation - application.js

Adds CSP header with style-src 'self' (strict, no unsafe-inline); adds /csp-report endpoint that logs any remaining violations to the server console

## Explanation - app.js

Removed redundant code which was previously used to display `react-tooltip` as part of the test harness toolbar. `react-tooltip` uses inline styles internally which was causing an error in the browser console. The test harness toolbar now uses the common canvas toolbar so `react-tooltip` is no longer needed.

tomlyn:
This branch not needed. Changes delivered in other branches. Closing.

PR #3256, opened 2026-07-23
#3228 Track removal of unsafe-inline from CSP and refactor dynamic in…

### LINKED ISSUE #3228: Track removal of unsafe-inline from CSP and refactor dynamic inline styles
Ref: 4374

tomlyn:
…line styles

Fixes: #3228 

 
           in the file; or

           it.


tomlyn:
Hi @matthoward366 and @caritaou 

I've been working on this issue [4374](https://github.ibm.com/NGP-TWC/wdp-abstract-canvas/issues/4374) which is a request to be able to run with a CSP (Content Security Policy) without needing to specify `style-src 'unsafe-inline'` in Elyra Canvas.

I changed the code to remove the inline styles in seven separate PRs which are referenced in [my comment](https://github.ibm.com/NGP-TWC/wdp-abstract-canvas/issues/4374#issuecomment-223986995) in the issue. This PR is the final piece of the puzzle (hopefully)

A new toggle has been added to the test harness toolbar that allows the user to switch between using the CSP or not. It does complicate the test harness code so I can remove that if you think it is not useful -- should we just have an environment variable instead that can be used to switch modes?

Here's Bob's explanation of what we are doing in this PR, edited by me:

## Content Security Policy support

Adds opt-in Content Security Policy enforcement to the test harness and threads a `cspNonce` through Common Properties so that runtime-style-injecting components work under a strict `style-src` policy.

---

### webpack.config.dev.cjs

The dev config previously used `style-loader`, which injects CSS at runtime by creating `<style>` tags via JavaScript DOM manipulation. Browsers treat these as inline styles and require `'unsafe-inline'` in `style-src` to permit them.

`style-loader` is replaced with `MiniCssExtractPlugin`, which extracts all CSS into real `.css` files served as static assets loaded via `<link rel="stylesheet">`. The browser fetches them as same-origin external resources, which are permitted under `style-src 'self'` with no `'unsafe-inline'` required.

---

### lib/csp-state.js (new file)

A small shared module holding the in-memory `cspEnabled` flag, which defaults to `false` so the harness starts with CSP off and users opt in via the toolbar toggle. Keeping the flag in its own module lets both `application.js` and `v1-ops-api.js` import it without creating a circular dependency.

---

### lib/application.js

Several changes are required to support toggleable CSP enforcement:

- **CSP middleware** — when CSP is enabled (via `csp-state.js`), a fresh cryptographic nonce is generated for every request and set in the `Content-Security-Policy` response header. The policy uses `style-src 'self' 'nonce-<nonce>'`; `'unsafe-inline'` is retained in `script-src` for the HMR dev toolchain. CSP violations are visible in the browser DevTools console.
- **Dynamic `index.html` serving** — `index.html` is now served dynamically (rather than as a static file) so the per-request nonce can be injected as `window.__CSP_NONCE__` before the page is sent to the browser. In dev mode the HTML is read from webpack's in-memory filesystem (populated by `HtmlWebpackPlugin`), and `_configureHmr` now returns a Promise that resolves only after the first build completes, ensuring the in-memory `index.html` exists before any browser request tries to read it.

---

### controllers/v1-ops-api.js

Adds a `POST /v1/test-harness/csp-enabled` endpoint that accepts `{ "enabled": true|false }` and updates the in-memory CSP flag in `csp-state.js`. The client navigates to a fresh page load after the call returns so the updated `Content-Security-Policy` header (or its absence) takes effect.

---

### src/client/App.js

- **CSP toggle** — a toggle in the harness toolbar calls the new `/v1/test-harness/csp-enabled` endpoint, then navigates to `window.location.origin + "/"` to trigger a genuine page reload so the new CSP header is applied. The toggle's initial state is derived from `window.__CSP_NONCE__`, which is injected by the server when CSP is active.
- **`cspNonce` forwarded to `CommonProperties`** — `window.__CSP_NONCE__` is passed as `cspNonce` in `propertiesConfig` so subcomponents that inject runtime styles can use a nonce-based permission.

---

### common-properties.jsx / expression.jsx / table components

Adds the `cspNonce` prop to the `propertiesConfig` PropTypes definition and threads it through every component that injects runtime styles:

- **expression.jsx** — passes the nonce to CodeMirror's `EditorView.cspNonce` extension so the expression editor can inject its required stylesheets under a strict `style-src` policy.
- **abstract-table.jsx, list.jsx, selectcolumns.jsx, someofselect.jsx, field-picker.jsx, expression-select-field-function.jsx** — each passes `cspNonce` down to `FlexibleTable` when constructing their table instances (three separate `FlexibleTable` instances in the expression builder's field, value, and function tables).
- **flexible-table.jsx** — adds `cspNonce` to its PropTypes and forwards it to `VirtualizedTable`.
- **virtualized-table.jsx** — passes the nonce to `AutoSizer` (react-virtualized) and `Draggable` (react-draggable) via their respective `nonce` props, covering all third-party components in the table stack that inject `<style>` tags at runtime.

---

### svg-canvas-renderer.js

Removes the `force-cache` option from both `d3.svg()` calls. Using a forced cache means that a previously cached SVG containing inline `style` attributes or `<style>` blocks will continue to be served from cache even after the server-side asset is cleaned up, causing CSP violations. Without `force-cache` the browser revalidates as normal.

---

## Documentation

- **docs/pages/02.02-content-security-policy.md** (new page) — documents the full requirements for running Elyra Canvas under a `style-src` policy without `'unsafe-inline'`: how to pass a nonce to `CommonProperties`, SVG asset constraints, the cache-busting consideration during CSP rollout, and which canvas inline-style APIs still require `'unsafe-inline'`.
- **docs/pages/04.08-properties-config.md** — documents the new `cspNonce` field in the Common Properties configuration, including a step-by-step usage example.
- **docs/pages/02-set-up.md** / **docs/mkdocs.yml** — registers the new CSP page in the documentation navigation.

caritaou (dismissed):

caritaou (approved):

matthoward366 (approved):
Looks good

PR #3258, opened 2026-07-24
Context for AI agents

sojinantony01:
# Add `context/` — AI-assisted quickstart for @elyra/canvas consumers

## The problem 

I picked up `@elyra/canvas` and thought — *how hard can it be?*

Reader, it was hard. Not because the library is bad — the docs are great. There are 50+ pages of them. Which is exactly the problem when you just want a network diagram with custom node colors and instead you're on page 34 reading about `enableStraightLinksAsFreeform` deprecation notices.

## The fix

A `context/` folder — paste it into any AI chat (BOB, ChatGPT, Claude, Cursor, Copilot), describe what you want, get working code in minutes instead of hours. i made below canvas in minutes and Agent was able to update/fix issues based on my use case and was able to identify bugs faster than before(I am not sharing my actual internal project diagram 😄)

Screenshot 2026-07-24 at 11 35 24 PM

Sample prompt i used
```

Add an Elyra canvas to this application using the context files in context/README.md.

Canvas requirements:
One Source node (pre-placed), followed by multiple Filter nodes, then multiple Transform nodes, then one Output node — all connected with one-directional arrows
Arrows must be colored
Nodes must use custom SVG icons with no background shape
All toolbar items enabled (undo, redo, cut, copy, paste, delete, comment, arrange, zoom, select all)

Behavior:
Every canvas action — including node creation, linking, deletion, move, rename, undo, and redo — must fire a callback
Every canvas change must be synced to React state, and state changes must re-render the canvas accordingly
Input port cardinality must be unlimited (max: -1) so multiple links can connect to the same node without being silently rejected
```

What's inside:
- **`starter-palette.json` + `starter-flow.json`** — 3 pre-connected nodes, inline SVG icons, zero server setup. Drop them in your project and you have a working canvas immediately.
- **Canvas type selector** — answer 5 questions, get the exact config for your use case
- **Copy-paste patterns** for nodes, links, colors, decorations, toolbar, callbacks, undo/redo, panels
- **Section-by-section links** to the official docs — no duplication, just the right pointer at the right time

## What this is NOT

A replacement for the docs. Every section links directly to the relevant official doc page. It's an on-ramp, not a detour.

---

> If this saves even one developer from reading the `enableStraightLinksAsFreeform` deprecation notice before they've written their first node, it was worth it.

 
           in the file; or

           it.


sojinantony01:
I have updated the folder path as suggested by @samueldmeyer 
CC: @tomlyn

PR #3280, opened 2026-07-30
#3279 Expression builder key nav

### LINKED ISSUE #3279: Expression builder keyboard navigation does not go to the single row radio selection
#4413

michaelcheungkm:
Fixes: #3279
 
           in the file; or

           it.


canvas_modules/common-canvas/src/tooltip/tooltip.jsx
```diff
@@ -587,14 +588,16 @@ ToolTip.propTypes = {
 	delay: PropTypes.number,
 	showToolTipOnClick: PropTypes.bool,
 	hoverable: PropTypes.bool, // If true, mouse cursor can be hovered over to the tooltip, instead of immediately disappearing.
+	alwaysTabbable: PropTypes.bool, // Set to true to make the trigger a tab stop even when the content isn't truncated
```
caritaou: We shouldn't need this. We follow the Carbon accessibility guidelines for tables https://carbondesignsystem.com/components/data-table/accessibility/ and only interactive elements should be tabble. In this case, only cells that are truncated that has a tooltip. The rest should be skipped

Image

caritaou (changes requested):

canvas_modules/common-canvas/src/tooltip/tooltip.jsx
michaelcheungkm: Oh ok I see, I will make it only tab through the bullet and the plus.

michaelcheungkm:
https://github.com/user-attachments/assets/e2e11829-3134-4245-8745-e4f2f331ea5d

@caritaou I made the changes in latest commit, works like this now.

edit: I forgot about the truncated part, making change

michaelcheungkm:
https://github.com/user-attachments/assets/074ceec2-de35-4426-8e3d-e6b255f6726d

There is this preexisting issue from before where it will sometimes just tab into nothing, I cannot quite figure out what is causing it yet. It seems to always happen 8 rows down from where you start tabbing from then it will cause the tab to always go somewhere not visible for 2 tabs on every row.

edit: probably to do with virtualized table but I can't find a fix for this

caritaou (approved):

PR #3304, opened 2026-08-10
#3303 Fix eslint-a11y warnings for common-canvas component

### LINKED ISSUE #3303: eslint-a11y warnings for common-canvas component
Ref: 4400

jsochava:
Fix: #3303 

### **Note**
Four files needed an eslint-disable because the code was already accessibility correct and the warning to my understanding was a limitation of the linter's static aria role table and not an actual bug:
- `cc-bottom-panel.jsx` / `cc-right-flyout.jsx` - these are "window splitter" drag handles, now given role="separator". aria query classifies separator as non interactive so the mouse handler on it still trips the linter even though a draggable separator is the standard accessible pattern.
- `cc-contents.jsx`- role="application" hands all keyboard handling to this canvas widget. aria-query still files application under "non-interactive," so both rules false positive here.
- `cc-text-toolbar.jsx` - this div is a positioning/focus management wrapper only; the actual toolbar behavior is provided by the nested `<Toolbar>` it wraps so there's no role to add to the wrapper itself that would be accurate.

 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/locales/common-canvas/locales/en.json
+++ b/canvas_modules/common-canvas/locales/common-canvas/locales/en.json
@@ -9,6 +9,8 @@
   "edit.pasteSelection": "Paste",
   "canvas.label": "Flow Editor",
   "canvas.contents": "Flow editor contents",
+  "canvas.bottomPanelResize": "Resize bottom panel",
+  "canvas.rightFlyoutResize": "Resize right flyout panel",
   "canvas.addComment": "New comment",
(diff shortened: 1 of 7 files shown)


matthoward366 (commented):
are you able to add comments above where the linting was disabled so we know why it was disabled?

matthoward366 (approved):
lgtm.  @caritaou can you have a quick look as wel?

caritaou (approved):
lgtm

PR #3321, opened 2026-08-13
Fix eslint-a11y warnings for palette component

### LINKED ISSUE #3315: eslint-a11y warnings for palette component
Ref #4404

amelial9:
Fixes #3315 

- The "Show more"/"Show less" toggle under truncated descriptions in the
  flyout search results was a `<div>` and could not be reached by keyboard.
  It is now a `<button>`, with click and Enter/Space stopped from
  propagating to the parent palette item, which would otherwise have added a
  node to the canvas at the same time.
- `.palette-list-item-desc-button` gains the usual reset (`background`,
  `border`, `font`, `text-align`, `display`) so the new `<button>` still
  renders as the inline link it replaced.
- Categories and tiles now carry `role="button"`, a `tabIndex` and an
  `onKeyDown` for Enter/Space, with `aria-disabled` and `tabIndex={-1}`
  keeping disabled tiles out of the tab order. Categories also gain
  `aria-current`, since the selected one was previously indicated by font
  weight alone and so was invisible to a screen reader.
- Node tips now appear on focus as well as hover, in both the dialog grid and
  the flyout search results.
- The "No results found" message is now `role="status"`, so a screen reader
  announces an empty search immediately rather than the user having to
  navigate to the message to discover it.
- Tabbing out of an empty result list called `tabOutOfOfPalette` unguarded —
  an optional prop that common-canvas never supplies. Now guarded, matching
  `tabOut` in `palette-content-list-item`.
- The truncated-results message carried an `onKeyDown` it could never fire,
  having no `tabIndex` and only text children. Removed.

Warnings suppressed:

- The dialog title bar and window edges handle dragging, maximizing and
  resizing the floating window. Those are pointer gestures with no keyboard
  equivalent rather than controls — giving them `role="button"` would announce
  a button that does nothing and wrap the real toolbar buttons in a fake one.
- The flyout category title div is asked for an `onFocus` beside its
  `onMouseOver`, but it cannot receive focus. Focus is handled on the
  `AccordionItem` it is passed to as its title, which ESLint does not look at.
  Confirmed in the harness — tabbing to a category does show the tip.
- The `tabIndex` on the no-results message is deliberate and is required by
  the public `tabOutOfOfPalette` prop, so it stays.
  
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/palette/palette-content-list-item-btn.jsx
+++ b/canvas_modules/common-canvas/src/palette/palette-content-list-item-btn.jsx
@@ -17,17 +17,42 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { injectIntl } from "react-intl";
+import KeyboardUtils from "../common-canvas/keyboard-utils.js";
 import defaultMessages from "../../locales/palette/locales/en.json";
 
 class PaletteContentListItemBtn extends React.Component {
+	constructor(props) {
+		super(props);
+
+		this.onClick = this.onClick.bind(this);
+		this.onKeyDown = this.onKeyDown.bind(this);
+	}
+
+	onClick(evt) {
+		// Stop the click reaching the parent palette list item which would
+		// otherwise add a node to the canvas.
+		evt.stopPropagation();
+		this.props.onClick(evt);
+	}
+
+	onKeyDown(evt) {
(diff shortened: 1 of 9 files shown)


tomlyn (approved):
LGTM

PR #3325, opened 2026-08-19
elyra-ai#4433 Fix CommonProperties crash from FlexibleTable

### LINKED ISSUE #3324: Dropdown causes CommonProperties to crash on fast change
Ref: 4433

nathanyu8th:
Fixes: #3324 

https://github.com/user-attachments/assets/52ae1272-5991-49fc-8463-b5273e92eca8

The properties panel was crashing because its shared table component recalculated column widths on every redraw and then saved the result even when the widths hadn't changed and each save triggered another redraw. In the running app, 95% of those saves stored a value that was already there. React gives up after 50 redraws triggered from inside a redraw, which is what produced the crash. A second, related bug made hidden tables report a width of zero, so they collapsed all columns to the 40px minimum and sprang back when shown. The dropdown was never at fault: changing any value redraws every control, and all eight places that use the table hand it a brand-new column list each time, so the table always thought something had changed. The fix is two guards in one file, only save widths when they actually differ, and ignore a width of zero, matching the pattern three neighboring methods in that same file already use.
 
           in the file; or

           it.

### DIFF
--- a/canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
@@ -182,7 +182,11 @@ class DropDown extends React.Component {
 		if (this.props.control.controlType === ControlType.SELECTCOLUMN) {
 			value = PropertyUtils.fieldStringToValue(value, this.props.control, this.props.controller);
 		}
-		this.props.controller.updatePropertyValue(this.props.propertyId, value);
+		// Carbon fires onSelectedItemChange from a downshift effect that re-runs when the selected
+		// item is re-synced, so dispatching an unchanged value re-enters this handler via the store.
+		if (!isEqual(value, this.props.value)) {
+			this.props.controller.updatePropertyValue(this.props.propertyId, value);
+		}
 	}
 
 	handleComboOnChange(evt) {
--- a/canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
+++ b/canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
@@ -182,9 +182,10 @@ class DropDown extends React.Component {
 		if (this.props.control.controlType === ControlType.SELECTCOLUMN) {
 			value = PropertyUtils.fieldStringToValue(value, this.props.control, this.props.controller);
 		}
-		// Carbon fires onSelectedItemChange from a downshift effect that re-runs when the selected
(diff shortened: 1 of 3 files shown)


canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -182,7 +182,11 @@ class DropDown extends React.Component {
 		if (this.props.control.controlType === ControlType.SELECTCOLUMN) {
 			value = PropertyUtils.fieldStringToValue(value, this.props.control, this.props.controller);
 		}
-		this.props.controller.updatePropertyValue(this.props.propertyId, value);
+		// Carbon fires onSelectedItemChange from a downshift effect that re-runs when the selected
+		// item is re-synced, so dispatching an unchanged value re-enters this handler via the store.
+		if (!isEqual(value, this.props.value)) {
```
caritaou: Instead of using props.value in this comparison, can you get the value from the store? Same as what is done for `handleOnInputChange()`

canvas_modules/common-canvas/src/common-properties/reducers/properties.js
```diff
@@ -23,7 +24,12 @@ function properties(state = {}, action) {
 		if (propertyId === null) {
 			return state;
 		}
-		var newState = state;
+		// A no-op write must not create a new state identity: every notification forces a
+		// SyncLane re-render, so re-setting a value to what it holds sustains an update loop.
+		if (typeof propertyId.row === "undefined" && isEqual(state[propertyId.name], action.property.value)) {
```
caritaou, line 29: A question from Bob:
> The most puzzling aspect of the last commit is that it reverts the FlexibleTable guards from commit 1 while solving the same problem a different way (via the reducer). The commit message says the reducer fix made them redundant, but given that columnWidths is local state (not Redux), the _setColumnWidths loop is an independent re-render churn path that the reducer can't prevent. You may want to ask the author why those guards were dropped.

caritaou: Basically we should still keep the changes from your first commit, those are valid checks

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
caritaou: @nathanyu8th One last thing, just to cover all paths, can you add a similar check here?

caritaou (approved):

PR #3345, opened 2026-09-15
#3344 Not possible to set breakpoints in the Chrome debugger

### LINKED ISSUE #3344: Not possible to set breakpoints in the Chrome debugger
Ref: 4437

tomlyn:
Fixes: #3344 
 
           in the file; or

           it.


tomlyn:
Hi @matthoward366  -- Bob and I finally figured out some changes that will enable the debugger in Chrome to work again with Elyra Canvas -- it works for both CSP On and Off. Can you review please?

canvas_modules/harness/webpack.config.dev.cjs
```diff
@@ -126,10 +125,20 @@ const plugins = [
...
-	// generates the source maps used for debugging.  Used instead of `devtool` option
+	// generates the source maps used for debugging.  Used instead of `devtool` option.
+	// moduleFilenameTemplate normalizes paths so that all source files (including those
+	// outside the harness directory) appear as ./-relative paths under webpack://,
+	// which Chrome DevTools can resolve correctly for breakpoints.
 	new webpack.SourceMapDevToolPlugin({
-		module: true,
-		columns: false
+		filename: "[file].map",
+		moduleFilenameTemplate: (info) => {
```
matthoward366, line 135: Is this needed specifically because of the CSP enablement?  It seems odd we need to add this.  Is this because of how we link to common-canvas?

tomlyn:
Three problems in the webpack dev config were preventing Chrome DevTools from working correctly:

1. Source map emitted inline instead of as a separate file

    SourceMapDevToolPlugin had no filename option, causing the entire source map to be base64-encoded and embedded directly inside canvasharness.js. This made the bundle very large and prevented the browser from re-fetching an updated map after a page reload.

    Fix: Added filename: "[file].map" so the map is emitted as a separate in-memory file served at /js/canvasharness.js.map.

2. Source map paths unresolvable by Chrome

    output.devtoolModuleFilenameTemplate: "[resource]" produced paths like webpack:///../src/palette/palette-flyout-content.jsx — a relative path with ../ that Chrome cannot resolve to an actual source file. Additionally, resolve.symlinks was not disabled, causing webpack to resolve the same source file via two different paths, producing duplicate and inconsistent source map entries.

    Fix: Replaced the output template with a moduleFilenameTemplate function on SourceMapDevToolPlugin that uses fs.realpathSync to canonicalise each path and then computes it relative to canvas_modules/, producing clean webpack:///./common-canvas/src/... and webpack:///./harness/src/... paths that Chrome resolves correctly. Added resolve.symlinks: false to prevent the duplicate path issue.

3. Some lines not breakpointable

    SourceMapDevToolPlugin had columns: false, which omits column-level position data from the map. Chrome needs both line and column information to determine which expressions within a line are breakpointable — without columns, many lines appear greyed out.

    Fix: Removed columns: false, restoring full column precision.

canvas_modules/harness/webpack.config.dev.cjs
tomlyn: I added an explanation of the changes as a comment to this PR: https://github.com/elyra-ai/canvas/pull/3345#issuecomment-5699206882

tomlyn:
I moved the require for 'fs' up to the top of the file and committed that change.

matthoward366 (approved):
lgtm

Issue #3351, opened 2026-09-18
Not compatible with Carbon React most recent version (1.116.0)


tomlyn:
Ref: 4443

PR #3352, opened 2026-09-18
#3351 Not compatible with Carbon React most recent version (1.116.0)

tomlyn:
Fixes: #3551
 
           in the file; or

           it.


tomlyn:
@matthoward366 @caritaou 

# Branch Changes: Carbon React 1.116.0 Compatibility

## Why it's needed

Carbon React `1.116.0` tightened its prop type validation so that label props (e.g. `labelText`, `titleText`, `legendText`) on its components **must be plain strings** — passing a React element (JSX) as a label is no longer accepted. Previously, common-properties controls passed their entire `controlItem` prop (a React element containing a styled label, required indicator, and tooltip icon) directly into Carbon components via these props. This broke with the new version.

---

## What changed

### Package upgrade (`package.json`)

`@carbon/react` bumped from `1.109.0` → `1.116.0`.

### Test infrastructure (`jest-setup.js`, `globalMock.js`)

The Carbon `enable-v12-dynamic-floating-styles` feature flag is enabled in both the Jest setup file and the global mock, as required by the new Carbon version's floating UI behaviour in tests.

### All input controls — the same pattern applied across 11 controls

The core fix applied consistently to `textfield`, `textarea`, `numberfield`, `passwordfield`, `slider`, `dropdown`, `multiselect`, `datefield`, `timefield`, `datepicker`, and `datepicker-range`:

- The Carbon label prop now receives only a plain string (the label text).
- `hideLabel` is set unconditionally so Carbon suppresses its own label rendering.
- `controlItem` is rendered as a separate sibling `<div>` above the Carbon component (skipped when inside a table cell).

### CSS fixes to restore correct visual layout

After moving `controlItem` out of Carbon's label prop and into a separate div, several controls needed CSS corrections to maintain their original appearance.

| File | Fix |
|---|---|
| `control-item.scss` | Sets `.properties-label-container` to `display: block` and `.tooltip-container` to `display: inline-flex; vertical-align: middle` so that the label text, `(required)` indicator, and info tooltip wrap naturally as inline content without flex stretch. |
| `textfield.scss` | Positions `cds--text-input__label-wrapper` (character counter) at the top-right aligned with the bottom of multi-line or single-line label areas. |
| `textarea.scss` | Same treatment for `.properties-textarea` / `cds--text-area__label-wrapper`. |
| `checkboxset.scss` | Hides the now-redundant `legend.cds--label` that Carbon still renders inside `CheckboxGroup`. |
| `datepicker-range.scss` | Adds a flex `properties-datepicker-range-label-container` to lay the two date-range sub-labels (Start / End) side by side, and adds the missing `@use` for Carbon tokens. |
| `numberfield.scss` | Wraps `NumberInput` + the optional number-generator button in a `properties-numberfield-input-container` flex row, replacing the previous `inline-flex` on the outer control that no longer works once the label is separated out. |

### Radioset — tooltip placement fix

The info tooltip icon that appears alongside each radio option's label was previously wrapped *inside* the `labelText` prop as a JSX element. It is now rendered as a sibling element after the `<RadioButton>`, keeping it visually adjacent without passing JSX into a label prop.

### Slider — `hideTextInput` prop added

`hideTextInput={false}` is now explicitly passed to Carbon's `Slider` to preserve the numeric input field that Carbon `1.116.0` hides by default.

tomlyn:
@caritaou 
Testing I did for this:

* Jest test run successfully
* Cypress tests run successfully
* I compared the output for the ‘properties’ page (accessed by clicking the ‘Common Properties documentation’ link) in both the changed version and the cloud version. They compared OK with just some very very minor vertical spacing differences.
* I manually tested many of the test harness paramDef files and compared them to the test harness running on the cloud and fixed anything that was different. I focused on those that Bob told me were most likely affected by these changes. Some things I observed were:
    + There are occasionally some small vertical spacing differences. Some of these seem to be affected by changing the @carbon/react version.
    + In Conditions_paramDef.json in the `Tab and SubTab` section where the left and right arrow buttons for the tabs and sub-tabs behave differently but it seemed like the new code was better  -- so I left it alone.
    + In action_paramDef.json, in the Alignment tab, the layout of the label + (required) + icon + character counter is slightly different but I think the new layout is better -- so I left it alone.
    + In datefield_paramDef.json, the position of one of the `i` icons has changed but, again, the way it looks now seems consistent with the other controls so I left it alone.
    + In defaults_paramDef.json, the position of some of the `i` icons has changed but, again, the way it looks now seems consistent with the other controls so I left it alone.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
```diff
@@ -306,7 +306,8 @@ class DropDown extends React.Component {
 					onChange={this.handleComboOnChange}
 					onInputChange={(evt) => this.handleOnInputChange(evt, dropDown.options)}
 					translateWithId={(id) => listBoxMenuIconTranslationIds[id]}
-					titleText={this.props.controlItem}
+					titleText=""
+					size="md"
```
caritaou: The ComboBox is appearing to be bigger than it was

Image

canvas_modules/common-canvas/src/common-properties/controls/multiselect/multiselect.scss
```diff
@@ -22,3 +22,8 @@
 		}
 	}
 }
+.cds--multi-select__wrapper {
+	.cds--multi-select {
+		height: 40px;
```
caritaou: How come this is needed? Does the `size` not work?

tomlyn: As discussed, I added a comment for this workaround.

canvas_modules/common-canvas/src/common-properties/controls/dropdown/dropdown.jsx
tomlyn: As discussed, I added a workaround for combo-box like the one for mult-iselect that overrides the block-size. I put the code in `dropdown.scss` because there doesn't seem to be a SCSS file specifically for combo-box.

caritaou (approved):
lgtm

